PR #1, opened 2026-02-06
Add licenses

nps-gattaca:


0w3n-d (approved):
We should also add something like `license = "MIT OR Apache-2.0"` to the Cargo.toml

PR #5, opened 2026-02-07
Allow for base dir specification of where shmem and persistence is saved

louisponet:
## Add configurable `base_dir` to shared memory and persistence paths

  All directory helpers previously hardcoded the base directory via `BaseDirs::data_dir()`. This PR threads an explicit `base_dir` parameter through the stack so callers can override where
  data lives.

  ### Changes

  - **`flux-utils/directories`** — `local_share_dir()` now returns just the base path (no app_name). Added `*_with_base` variants for all directory helpers. Changed `app_name` bounds from
  `AsRef<str>` to `AsRef<Path>`.

  - **`flux-communication`** — Added `shmem_queue_with_base_dir`, `ShmemData::open_or_init_with_base_dir`, `Timer::new_with_base_dir`. Originals remain as thin wrappers using
  `local_share_dir()`. Also: stale shmem files are now removed and retried instead of panicking.

  - **`flux/persistence`** — Added `persist_in_base_dir`, `persist_dir_with_base_dir`, `load_with_base_dir` to `Persistable`. `PersistingQueueTile` now stores a `base_dir` field.

  - **`flux/spine`** — New required trait method `fn new_in_base_dir(base_dir: impl AsRef<Path>) -> Self` on `FluxSpine`.

  - **`spine-derive`** — Generated spine structs carry a `base_dir: PathBuf` field. Macro generates `new_with_base_dir` and uses `_with_base_dir` variants for all shmem/data creation.

  ### Notes

  Non-breaking additive change. All existing APIs keep their defaults via `local_share_dir()`.

### DIFF
--- a/crates/flux-communication/src/timer.rs
+++ b/crates/flux-communication/src/timer.rs
@@ -1,4 +1,4 @@
-use std::{collections::HashMap, fmt::Display, sync::Mutex};
+use std::{collections::HashMap, fmt::Display, path::Path, sync::Mutex};
 
 use flux_timing::{Duration, Instant, InternalMessage, Nanos};
 use once_cell::sync::Lazy;
@@ -60,7 +60,7 @@ pub struct Timer {
 }
 
 impl Timer {
-    pub fn new<A: AsRef<str>, S: Display>(app_name: A, name: S) -> Self {
+    pub fn new<A: AsRef<Path>, S: Display>(app_name: A, name: S) -> Self {
         let dirstr = shmem_dir_queues_string(&app_name);
         let _ = std::fs::create_dir_all(&dirstr);
 
--- a/crates/flux-communication/src/timer.rs
+++ b/crates/flux-communication/src/timer.rs
@@ -1,12 +1,10 @@
 use std::{collections::HashMap, fmt::Display, path::Path, sync::Mutex};
 
 use flux_timing::{Duration, Instant, InternalMessage, Nanos};
+use flux_utils::directories::{local_share_dir, shmem_dir_queues_with_base};
 use once_cell::sync::Lazy;
 
-use crate::{
-    queue::{Producer, Queue, QueueType},
-    shmem_dir_queues_string,
-};
+use crate::queue::{Producer, Queue, QueueType};
 
 /// A single timing interval measured on one machine.
 ///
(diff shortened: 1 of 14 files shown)


crates/flux-communication/src/timer.rs
```diff
@@ -60,8 +58,16 @@ pub struct Timer {
 }
 
 impl Timer {
-    pub fn new<A: AsRef<str>, S: Display>(app_name: A, name: S) -> Self {
-        let dirstr = shmem_dir_queues_string(&app_name);
+    pub fn new<A: AsRef<Path>, S: Display>(app_name: A, name: S) -> Self {
+        Self::new_with_base_dir(local_share_dir(), app_name, name)
+    }
+
+    pub fn new_with_base_dir<D: AsRef<Path>, A: AsRef<Path>, S: Display>(
```
gd-0, line 65: Shouldnt we deprecate all use of local dir and use base_dir from spine?

crates/flux/src/persistence/persisting_tile.rs
```diff
@@ -12,18 +15,27 @@ pub const TIMESTAMP_FORMAT_UTC: &str = "%Y-%m-%d_%H:%M_utc";
 pub struct PersistingQueueTile<T: Persistable> {
     pending_data: Vec<InternalMessage<T>>,
     last_persist_time: Nanos,
+    base_dir: PathBuf,
 }
 
 impl<T: Persistable> PersistingQueueTile<T> {
     pub fn new() -> Self {
-        Self { pending_data: Vec::new(), last_persist_time: Nanos::now() }
+        Self::new_with_base_dir(local_share_dir())
```
gd-0, line 23: Again same as above. Otherwise can't we have half the data appearing in local/share. Half in whatever base dir we set. 

Can't we make the base dir default to local/share if spine is made without a base_dir and then if it is set. Use whatever the user sets

crates/flux-communication/src/timer.rs
louisponet: I think the idea is to be flexible? I did not check the timers in the adapter yet actually so might not be using the right base dir there yet.

louisponet: Just double checked that indeed the timers used by the spine consumers are not using the right dir yet. Will make that change.

louisponet: Just double checked that indeed the timers used by the spine consumers are not using the right dir yet. Will make that change.

gd-0 (approved):

PR #6, opened 2026-02-08
fix Duration overflows

ltitanb:


crates/flux-timing/src/global_clock.rs
niki4smirn:
```suggestion
pub(super) const MULTIPLIER: u128 = 100_000_000;
```

niki4smirn:
I'd also add a few tests to prevent similar issues in the future

louisponet:
Don't love the conversion back and forth to u128, why not just have different scales for different functions? Surely you don't need secs to be as precise as nanos etc?

ltitanb:
updated @louisponet @niki4smirn

niki4smirn (approved):

PR #11, opened 2026-02-10
remove wincode_ser_into_slice

niki4smirn:
write_or_enqueue_with now accepts Vec<u8> and forwards it as a slice with len = vec.len(), not vec.capacity()


gd-0 (approved):

gd-0:
I think we might be able to remove the wincode featuer flag from the network crate now

PR #20, opened 2026-02-12
implement wincode::SchemaRead and SchemaWrite for ArrayStr

ninaiiad:


gd-0 (changes requested):
ArrayStr is purely a wrapper around ArrayVec is there no way to tell wincode to just use the inner impl?

e.g., think it has something called Pod

crates/flux-utils/src/arrayvec.rs
```diff
@@ -447,6 +447,13 @@ impl<const N: usize> core::fmt::Debug for ArrayStr<N> {
     }
 }
 
+impl<const N: usize> core::hash::Hash for ArrayStr<N> {
+    #[inline]
+    fn hash<H: core::hash::Hasher>(&self, state: &mut H) {
+        self.as_str().hash(state);
```
gd-0: should this not be the same impl as arrayvec? so no as_str()

gd-0 (approved):

PR #24, opened 2026-02-17
fix tcp reconnection handling

ninaiiad:
previously when the server finally reconnected to an address it did not emit an Accept event and the user never received that connection's token

### DIFF
--- a/crates/flux-network/src/tcp/connector.rs
+++ b/crates/flux-network/src/tcp/connector.rs
@@ -30,7 +30,7 @@ pub enum ConnectionVariant {
 
 /// Event emitted by [`TcpConnector::poll_with`] for each notable IO occurrence.
 pub enum PollEvent<'a> {
-    /// A new inbound connection was accepted from a listener.
+    /// A new connection was accepted from a listener.
     ///
     /// - `listener`: token of the listening socket that accepted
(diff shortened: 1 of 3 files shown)


crates/flux-network/src/tcp/connector.rs
```diff
@@ -235,6 +238,7 @@ impl ConnectionManager {
                 return;
             }
             self.to_be_reconnected.swap_remove(i);
```
ltitanb: `swap_remove` returns already the element you need to push to `self.newly_connected`

crates/flux-network/src/tcp/connector.rs
```diff
@@ -414,6 +418,9 @@ impl TcpConnector {
         F: for<'a> FnMut(PollEvent<'a>),
     {
         self.conn_mgr.maybe_reconnect();
+        for (token, peer_addr) in self.conn_mgr.newly_connected.drain(..) {
+            handler(PollEvent::Accept { listener: token, stream: token, peer_addr });
+        }
```
ltitanb, line 423: why not pass the handler to maybe_reconnect so you avoid the additional vector push and drain

ltitanb (approved):

PR #26, opened 2026-02-18
panic on thread boost fail

BrycePy:
so we don't accidentally run latency critical tiles without a proper cpu affinity / thread priority


gd-0 (changes requested):
I don't think this is a good idea to be honest. 
Often when running locally this can fail and it's totally fine.

If you care enough about latency you just need to check the logs once and fix it

PR #38, opened 2026-03-04
Add collaborative consume of messages from queues

alexX512:
---

<a href="https://app.devin.ai/review/gattaca-com/flux/pull/38" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>


crates/flux-communication/src/queue/mod.rs
```diff
@@ -27,14 +27,18 @@ pub enum QueueType {
...
     pub elsize: usize,         // 16
     pub mask: usize,           // 24
-    pub count: AtomicUsize,    /* 32 */
+    pub count: AtomicUsize,    // 32
+    _pad2: [u8; 32],           // 64 — pad to end of cache line
 
-                               /* add queue hash mismatch */
+    // Cache line 1: written by collaborative consumers only
+    pub consume_cursor: AtomicUsize, // 64
+                                     // 56 bytes implicit padding (align 64 → total 128)
```
vladimir-ea: does the `consume_cursor` need to be in the `QueueHeader` at all? if it is used separately why not have a separate struct?

crates/flux-communication/src/queue/mod.rs
```diff
@@ -248,6 +258,36 @@ impl<T: Copy> InnerQueue<T> {
         }
     }
 
+    #[inline]
+    pub fn try_consume_collaborative(&self, el: &mut T) -> Result<(), ReadError> {
+        loop {
+            let cur = self.header.consume_cursor.load(Ordering::Acquire);
```
vladimir-ea: what if the queue has already wrapped before the first collaborative consumer joins? or what if it wraps between collaborative consumes? wouldn't that result in collaborative consumers skipping chunks of data?

crates/flux-communication/src/queue/mod.rs
```diff
@@ -27,14 +27,18 @@ pub enum QueueType {
...
     _pad1: [u8; 6],            // 8
     pub elsize: usize,         // 16
     pub mask: usize,           // 24
-    pub count: AtomicUsize,    /* 32 */
+    pub count: AtomicUsize,    // 32
+    _pad2: [u8; 32],           // 64 — pad to end of cache line
 
-                               /* add queue hash mismatch */
+    // Cache line 1: written by collaborative consumers only
+    pub consume_cursor: AtomicUsize, // 64
```
louisponet: We need to keep track of multiple consumers with labels, let's add 256 slots with some ArrayString<20> for a label and the current pos for each

crates/flux-communication/src/queue/mod.rs
louisponet: The idea is that this can be tracked by multiple processes and also used to reattach where someone left off in the case of something going wrong

crates/flux-communication/src/queue/mod.rs
louisponet: I think that's desired behaviour. Still no backpressure as we don't want the whole system to die because 1 misbehaving consumer

crates/flux-communication/src/queue/mod.rs
```diff
@@ -248,6 +258,36 @@ impl<T: Copy> InnerQueue<T> {
...
+    pub fn try_consume_collaborative(&self, el: &mut T) -> Result<(), ReadError> {
+        loop {
+            let cur = self.header.consume_cursor.load(Ordering::Acquire);
+            if cur >= self.count() {
+                return Err(ReadError::Empty);
+            }
+            if self
+                .header
+                .consume_cursor
+                .compare_exchange_weak(cur, cur + 1, Ordering::AcqRel, Ordering::Relaxed)
```
louisponet: why not fetchadd here to a local variable that is used until it's not empty? I also don't think we should spinloop?

alexX512: We should atomically increase and check that cursor is before producer position. Just with fetch_add it is possible that we do increase in 2 readers when we had only 1 element because `cur >= self.count()` was passed in both of them before increase

Also yeh as we use it only under tile which is spinloop it can be skipped

crates/flux-communication/src/queue/mod.rs
vladimir-ea: and why does that mean it must be part of the `QueueHeader`?

crates/flux-communication/src/queue/mod.rs
vladimir-ea: ok - so lets say the queue has 50 million elements in it and the collaborating consumer joins, do they consume 50 million messages before catching up?

crates/flux-communication/src/queue/mod.rs
vladimir-ea: > We need to keep track of multiple consumers with labels, let's add 256 slots with some ArrayString<20> for a label and the current pos for each

why multiple consumers? isn't the idea here that each message is consumed by exactly one consumer?

crates/flux-communication/src/queue/mod.rs
alexX512: Now in case we have a big gap we move cursor to the earliest available slot

crates/flux-communication/src/queue/mod.rs
```diff
@@ -498,13 +553,26 @@ impl<T: Copy> ConsumerBare<T> {
         }
     }
 
+    #[inline]
+    pub fn acquire_next_slot(&mut self) {
+        self.collaborative_slot =
+            unsafe { &*self.collaborative_cursor }.fetch_add(1, Ordering::Relaxed);
+        self.collaborative_version = self.queue.version_at(self.collaborative_slot);
+    }
```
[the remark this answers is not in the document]
alexX512: Basic tests considered to not use collaborative

crates/flux-communication/src/queue/mod.rs
```diff
@@ -498,13 +553,26 @@ impl<T: Copy> ConsumerBare<T> {
...
+
     #[allow(clippy::not_unsafe_ptr_arg_deref)]
     pub fn init_header(consumer_ptr: *mut ConsumerBare<T>, queue: Queue<T>) {
         unsafe {
             (*consumer_ptr).pos = queue.cur_pos();
             (*consumer_ptr).expected_version = queue.version();
             (*consumer_ptr).mask = queue.header.mask;
-            (*consumer_ptr).queue = queue
+            (*consumer_ptr).collaborative_slot = usize::MAX;
+            (*consumer_ptr).collaborative_cursor = queue.group_cursor("default");
```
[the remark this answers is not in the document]
alexX512: init_header not used in tests

crates/flux-communication/src/queue/mod.rs
```diff
@@ -498,13 +553,38 @@ impl<T: Copy> ConsumerBare<T> {
...
+    pub fn acquire_earliest_available_slot(&mut self) {
+        let collaborative_cursor = unsafe { &*self.collaborative_cursor };
+        let delta = self
+            .queue
+            .count()
+            .saturating_sub(collaborative_cursor.load(Ordering::Relaxed) + self.queue.len());
+
+        self.collaborative_slot = collaborative_cursor.fetch_add(delta, Ordering::Relaxed);
+        self.collaborative_version = self.queue.version_at(self.collaborative_slot);
+    }
```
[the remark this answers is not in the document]
alexX512: This is expected, it is ok in our case to skip some available slots because now we are already too far from head

crates/flux-communication/src/queue/mod.rs
```diff
@@ -448,15 +474,44 @@ impl<T> AsMut<Producer<T>> for Producer<T> {
...
-    is_running: u8,        // 25
-    _pad: [u8; 11],        // 36
-    queue: Queue<T>,       // 56 fat ptr: (usize, pointer)
+    pos: usize,                               // 8
+    mask: usize,                              // 16
+    expected_version: u64,                    // 24
+    is_running: u8,                           // 25
+    _pad: [u8; 7],                            // 32
+    collaborative_version: u64,               // 40
+    collaborative_slot: usize,                // 48 — absolute queue count of the next slot to read
```
louisponet: do we need this? Can't we just do the same logic as what we do with the pos now but with self.pos = fetchadd?

crates/flux-communication/src/queue/mod.rs
```diff
@@ -448,15 +474,44 @@ impl<T> AsMut<Producer<T>> for Producer<T> {
...
+    collaborative_cursor: *const AtomicUsize, // 56
+    queue: Queue<T>,                          // 72 fat ptr: (usize, pointer)
 }
 
+unsafe impl<T> Send for ConsumerBare<T> {}
+
 impl<T: Copy> ConsumerBare<T> {
+    pub fn new_basic_test(queue: Queue<T>) -> Self {
```
louisponet: let's add that maybe with a #[cfg(test)]

crates/flux-communication/src/queue/mod.rs
```diff
@@ -448,15 +474,44 @@ impl<T> AsMut<Producer<T>> for Producer<T> {
...
+        Self::new_test(queue, std::ptr::null())
+    }
+
+    pub fn new_test(queue: Queue<T>, cursor: *const AtomicUsize) -> Self {
```
louisponet: same as above, let's just ahve a impl block that is #[cfg(test)]

crates/flux-communication/src/queue/mod.rs
```diff
@@ -549,6 +622,14 @@ pub struct Consumer<T: 'static + Copy> {
 }
 
 impl<T: 'static + Copy> Consumer<T> {
+    pub fn new_collaborative_test(queue: Queue<T>, cursor: *const AtomicUsize) -> Self {
```
louisponet: same as above

louisponet (commented):
I'm not super loving that we don't just have uniform logic. I kind of want to see where broadcast consumers are in shmem as well

crates/flux-communication/src/queue/mod.rs
```diff
@@ -34,8 +57,10 @@ pub struct QueueHeader {
     pub mask: usize,           // 24
     pub count: AtomicUsize,    /* 32 */
 
-                               /* add queue hash mismatch */
+    group_labels: [[u8; GROUP_LABEL_LEN]; MAX_GROUPS],
```
louisponet: let's make this use ArrayStr or ArrayString?

louisponet (commented):
nice looks pretty much good to go. The only thing I'm not 100% sure about is the try_init each time on the consume or broadcast. I know it will be probably predicted away, but shouldn't we just do that on startup? Do we ever want both collaborative and broadcast consuming in 1 consumer?

alexX512:
> but shouldn't we just do that on startup? 

There are some problems with startup init:
* One issue is at which moment we will specify which queues are collaborative, so we should be careful to not create collaborative queue from spine that will not use it as it will acquire slot and it will be missed from real collaborative consumers
* Another one that we have limited number of labels and spine always call queue constructor even if it doesn't use it so it always will create a new label
* Also with lazy init it's easier to track who actually use the queue and spedPast cases

So as it will be always correctly branch predicted after init and simplify tracking of the above cases I would like to keep lazy init

> Do we ever want both collaborative and broadcast consuming in 1 consumer?

If you are talking about why do we need  broadcast/collab suffix it is mostly for debugging/observabillity of who read queue in collab mod if we will add it to overseer. Otherwise I don't think that consumer from the same thread will need both mods at the same time

louisponet (approved):

PR #39, opened 2026-03-05
DCache implementation

ninaiiad:
---

<a href="https://app.devin.ai/review/gattaca-com/flux/pull/39" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/crates/flux-utils/benches/dcache.rs
+++ b/crates/flux-utils/benches/dcache.rs
@@ -1,8 +1,8 @@
 use std::{
     hint::{black_box, spin_loop},
-    sync::{Arc, Barrier},
+    sync::Arc,
     thread,
-    time::{Duration, Instant},
+    time::{Duration, SystemTime, UNIX_EPOCH},
 };
 
 use criterion::{BenchmarkId, Criterion, Throughput, criterion_group, criterion_main};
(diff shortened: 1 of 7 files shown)


crates/flux-utils/src/datastore.rs
```diff
@@ -0,0 +1,267 @@
...
+    sync::atomic::{AtomicUsize, Ordering::*},
+};
+
+#[derive(Debug, Clone, Copy)]
+pub struct DataStoreRef {
+    pub offset: usize,
+    pub len: usize,
+}
+
+#[derive(Debug)]
```
louisponet: you can use thiserror::Error to get some niceties without needing manual impls

crates/flux-utils/src/datastore.rs
```diff
@@ -0,0 +1,267 @@
...
+/// Intended use: the caller exchanges [`DataStoreRef`]s through an external
+/// ordered channel (e.g. a spine queue) and the channel's Release/Acquire
+/// pair is the synchronisation point for the payload bytes stored here.
+pub struct DataStore<const N: usize> {
+    data: Box<UnsafeCell<[u8; N]>>,
+    reserved: AtomicUsize,
+}
+
+impl<const N: usize> DataStore<N> {
+    const _POW_2_CHECK: () = assert!(N.is_power_of_two());
```
louisponet: there's a library we use in the builder for this called const_assert, it's quite good maybe we can do it here too?

louisponet (commented):
what happens if someone was reading the data while another one overwrites it? Do we assume that will just not happen? I think that's ok but also quite dangerous.

crates/flux-utils/src/datastore.rs
```diff
@@ -0,0 +1,267 @@
...
+
+    pub fn read(&self, r: DataStoreRef, buf: &mut [u8]) -> Result<(), DataStoreError> {
+        if r.len > N {
+            return Err(DataStoreError::DataLenExceedsCapacity(r.len, N));
+        }
+        if r.len > buf.len() {
+            return Err(DataStoreError::BufferTooSmall(buf.len(), r.len));
+        }
+
+        if r.offset + N < self.reserved.load(Acquire) {
```
ninaiiad: replace this with "epoch" prefix

crates/flux-utils/src/datastore.rs
```diff
@@ -0,0 +1,276 @@
...
+
+impl<const N: usize> DataStore<N> {
+    const OFFSET: usize = std::mem::size_of::<usize>();
+
+    pub fn new() -> Self {
+        const { assert!(N.is_power_of_two() && N > Self::OFFSET) };
+        Self { data: Box::new(UnsafeCell::new([0u8; N])), reserved: AtomicUsize::new(0) }
+    }
+
+    pub fn read(&self, r: DataStoreRef, buf: &mut [u8]) -> Result<(), DataStoreError> {
```
ninaiiad: maybe this could also accept a closure that would "transform" that chunk into something and return that if stale checks on both ends succeed?
like the decoder in helix needs a read-only reference anyway, so why copy?
this should be safer than return a reference

ps: would not work with two-part cut payloads though?

gd-0 (commented):
can you write a bench for this and compare it vs crossbeam?

crates/flux-utils/src/dcache.rs
gd-0, line 1: should add inline tags to the fns

crates/flux-utils/src/datastore.rs
```diff
@@ -0,0 +1,276 @@
...
+
+        let mut offset = [0u8; size_of::<usize>()];
+        Self::ring_read(base, r.offset, &mut offset);
+        if usize::from_ne_bytes(offset) != r.offset {
+            return Err(DataStoreError::StaleData);
+        }
+
+        Self::ring_read(base, r.offset + Self::OFFSET, &mut buf[..r.len]);
+
+        if r.offset + N < self.reserved.load(Acquire) {
```
gd-0: shouldn't this be `<=` ? the value at `offset` should be included

crates/flux-utils/src/dcache.rs
gd-0, line 1: so something firedancer does which we are missing is to align writes to cachelines.

so imagine we have:
( [] is a cacheline)
[<msg1>, <first_byte_msg_2>] [<rest_of_msg_2>]

if reading msg_2 we need to load 2 cachelines. 

What firedancer does is that when finding the next index to start writing it will skip past byttes on the same cacheline until we hit a new one. That leaves some 0 bytes in the buffer but think that trade-off is worth it

gd-0, line 1: another point from firedancer. they make this single-producer. Wonder if that is preferable and we just attach which tile's dcache the offset is for 

We should bench both of these implementations (sp vs mp) to see what latency trade-off there is for 1 to 4 producers

crates/flux-utils/src/datastore.rs
```diff
@@ -0,0 +1,276 @@
...
+        let from = self.reserved.fetch_add(Self::OFFSET + src.len(), Release);
+        let base = self.data.get().cast::<u8>();
+        Self::ring_write(base, from, &from.to_ne_bytes());
+        Self::ring_write(base, from + Self::OFFSET, src);
+
+        Ok(DataStoreRef { offset: from, len: src.len() })
+    }
+
+    fn ring_write(base: *mut u8, from: usize, src: &[u8]) {
+        let ix = from % N;
```
gd-0: we make sure that length is a power of 2. so don't need to mod here. can & with len-1

crates/flux-utils/src/datastore.rs
```diff
@@ -0,0 +1,276 @@
...
+    }
+
+    pub fn write(&self, src: &[u8]) -> Result<DataStoreRef, DataStoreError> {
+        if src.len() + Self::OFFSET > N {
+            return Err(DataStoreError::DataLenExceedsCapacity(src.len(), N - Self::OFFSET));
+        }
+
+        let from = self.reserved.fetch_add(Self::OFFSET + src.len(), Release);
+        let base = self.data.get().cast::<u8>();
+        Self::ring_write(base, from, &from.to_ne_bytes());
```
gd-0: unclear to me why we write the offset into the message? don't we have this in the DataStoreRef? isn't this just wasted space?

crates/flux-utils/src/dcache.rs
```diff
@@ -0,0 +1,276 @@
+use std::{
+    cell::UnsafeCell,
+    sync::atomic::{AtomicUsize, Ordering::*},
+};
+
+#[derive(Debug, Clone, Copy)]
+pub struct DataStoreRef {
+    pub offset: usize,
+    pub len: usize,
```
gd-0, line 11: minor nit but we should u32 these

crates/flux-utils/src/datastore.rs
```diff
@@ -0,0 +1,276 @@
...
+
+        Ok(())
+    }
+
+    pub fn write(&self, src: &[u8]) -> Result<DataStoreRef, DataStoreError> {
+        if src.len() + Self::OFFSET > N {
+            return Err(DataStoreError::DataLenExceedsCapacity(src.len(), N - Self::OFFSET));
+        }
+
+        let from = self.reserved.fetch_add(Self::OFFSET + src.len(), Release);
```
gd-0: we're releasing here before we write which doesn't do anything. maybe unecassary as we force through another channel but we should be more explicit if so and just use Relaxed

crates/flux-utils/src/datastore.rs
ninaiiad: I don't think so - it's about the length of data that has been reserved rather than an index of its own and it is 0 at the start

crates/flux-utils/src/dcache.rs
ninaiiad: I don't think we can do this with the offset - this one overflows capacity, it's not a valid index, so making it u32 would mean the datastore can only process 4gb of data before this overflows and the staleness checks break

crates/flux-utils/benches/dcache.rs
ninaiiad, line 1: producer bench

with 1 producer crossbeam does better (twice as much) with both msg sizes - 
with 2 producers crossbeam's twice as good with 50kb messages but with 400kb a little worse than both mp and sp
with 4 producers and 50kb messages it's comparable and a little worse than mp, at 400kb it's noticeably worse than mp

consumer bench

I feel like it may not be super fair because consumer performance increases when there are more producers?
otherwise mp+sp do better with bigger messages with 1-2-4 producers and better on smaller messages only for 4 producers

run results samples:
```
producer/mp/2p_409600   time:   [28.717 µs 30.194 µs 32.026 µs]
                        thrpt:  [11.911 GiB/s 12.634 GiB/s 13.284 GiB/s]
                 change:
                        time:   [−6.6437% −1.2355% +5.0422%] (p = 0.69 > 0.05)
                        thrpt:  [−4.8002% +1.2510% +7.1165%]
                        No change in performance detected.
Found 3 outliers among 10 measurements (30.00%)
  1 (10.00%) low mild
  2 (20.00%) high severe
producer/sp/2p_409600   time:   [37.452 µs 38.652 µs 40.464 µs]
                        thrpt:  [9.4273 GiB/s 9.8694 GiB/s 10.186 GiB/s]
                 change:
                        time:   [−6.7436% −2.8825% +1.0381%] (p = 0.19 > 0.05)
                        thrpt:  [−1.0274% +2.9680% +7.2313%]
                        No change in performance detected.
producer/crossbeam/2p_409600
                        time:   [39.911 µs 41.354 µs 42.899 µs]
                        thrpt:  [8.8922 GiB/s 9.2244 GiB/s 9.5579 GiB/s]
                 change:
                        time:   [−4.4954% +6.4103% +22.429%] (p = 0.38 > 0.05)
                        thrpt:  [−18.320% −6.0241% +4.7070%]
                        No change in performance detected.
```

```
consumer/mp/2p_409600   time:   [13.863 µs 14.072 µs 14.286 µs]
                        thrpt:  [26.702 GiB/s 27.109 GiB/s 27.517 GiB/s]
                 change:
                        time:   [+1.5165% +8.4795% +18.648%] (p = 0.06 > 0.05)
                        thrpt:  [−15.717% −7.8166% −1.4939%]
                        No change in performance detected.
Found 1 outliers among 10 measurements (10.00%)
  1 (10.00%) high severe
consumer/sp/2p_409600   time:   [19.180 µs 19.496 µs 20.197 µs]
                        thrpt:  [18.888 GiB/s 19.566 GiB/s 19.889 GiB/s]
                 change:
                        time:   [−0.9704% +3.0178% +7.4864%] (p = 0.18 > 0.05)
                        thrpt:  [−6.9650% −2.9294% +0.9799%]
                        No change in performance detected.
consumer/crossbeam/2p_409600
                        time:   [26.685 µs 27.317 µs 27.950 µs]
                        thrpt:  [13.649 GiB/s 13.965 GiB/s 14.295 GiB/s]
                 change:
                        time:   [+42718% +43852% +44972%] (p = 0.00 < 0.05)
                        thrpt:  [−99.778% −99.772% −99.766%]
                        Performance has regressed.
```

crates/flux-utils/src/datastore.rs
```diff
@@ -0,0 +1,345 @@
...
+
+        if r.offset + N < self.reserved.load(Relaxed) {
+            return Err(DataStoreError::StaleData);
+        }
+
+        Ok(result)
+    }
+
+    #[inline]
+    pub fn write(&self, src: &[u8]) -> Result<DataStoreRef, DataStoreError> {
```
ninaiiad: same as with map - might be better to accept a function that serialises some data instead of a ready-made buffer

crates/flux-utils/src/datastore.rs
```diff
@@ -0,0 +1,345 @@
...
+            return Err(DataStoreError::StaleData);
+        }
+
+        self.copy(r.len, r.offset, buf);
+
+        // Prevent the compiler from sinking the data reads past the staleness
+        // check below.
+        compiler_fence(AcqRel);
+
+        if r.offset + N < self.reserved.load(Relaxed) {
```
gd-0: if we actually want synchronisation for staleness checks then we need to make self.reserved Acq. 
I don't think a pure compiler fence is enough here.

crates/flux-utils/src/datastore.rs
```diff
@@ -0,0 +1,345 @@
...
+/// synchronised multi-producer allocation and leaves read-write ordering and
+/// synchronisation to the caller.
+///
+/// Intended use: the caller exchanges [`DataStoreRef`]s through an external
+/// ordered channel (e.g. a spine queue) and the channel's Release/Acquire
+/// pair is the synchronisation point for the payload bytes stored here.
+pub struct DataStore<const N: usize> {
+    data: Box<UnsafeCell<[u8; N]>>,
+    reserved: AtomicUsize,
+}
```
gd-0: should add cacheline padding here. data and reserved should not share a cacheline

crates/flux-utils/src/dcache.rs
```diff
@@ -0,0 +1,345 @@
...
+/// Ring buffer data storage inspired by Firedancer's 'dcache'. Provides
+/// synchronised multi-producer allocation and leaves read-write ordering and
+/// synchronisation to the caller.
+///
+/// Intended use: the caller exchanges [`DataStoreRef`]s through an external
+/// ordered channel (e.g. a spine queue) and the channel's Release/Acquire
+/// pair is the synchronisation point for the payload bytes stored here.
+pub struct DataStore<const N: usize> {
+    data: Box<UnsafeCell<[u8; N]>>,
+    reserved: AtomicUsize,
```
gd-0, line 33: I don't love that we are thrashing this atomic. Might need a bit of a rethink

crates/flux-utils/benches/dcache.rs
gd-0: Especially for SP this seems quite off. I have a strong feeling there is something wrong with this bench

crates/flux-utils/src/datastore.rs
ninaiiad: I think these serve different purposes

crates/flux-utils/src/dcache.rs
ninaiiad: if we find that having a dcache per producer is good enough for our purpose then we can do without the atomic

crates/flux-utils/src/dcache.rs
```diff
@@ -0,0 +1,377 @@
...
+    #[inline]
+    pub fn read(&self, r: DCacheRef, buf: &mut [u8]) -> Result<(), DCacheError> {
+        if r.len > N {
+            return Err(DCacheError::DataLenExceedsCapacity(r.len, N));
+        }
+        if r.len > buf.len() {
+            return Err(DCacheError::BufferTooSmall(buf.len(), r.len));
+        }
+
+        if r.offset + N < self.reserved.load(Relaxed) {
```
louisponet: I think we should use the offset that we find in the data here rather than reading the self.reserved. I think in general we want to avoid reading the same data in the consumer that the writer uses during operation. That's also why in the queue we read the seq in teh Seqlock + expected version to decide what to do rather than directly read the count variable itself

crates/flux-utils/src/dcache.rs
```diff
@@ -0,0 +1,377 @@
...
+            return Err(DCacheError::StaleData);
+        }
+
+        self.copy(r.len, r.offset, buf);
+
+        // Prevents the sinking of the data reads past the staleness
```
louisponet: rather than this fence etc, can't we just reread the offset byte after reading the full data to double check that nothing overwrote it? A bit similar to a seqlock

crates/flux-utils/src/dcache.rs
```diff
@@ -0,0 +1,377 @@
...
+            std::ptr::copy_nonoverlapping(base.add(from_ix), buf.as_mut_ptr(), head_len);
+            let tail_len = r_len - head_len;
+            if tail_len > 0 {
+                std::ptr::copy_nonoverlapping(base, buf[head_len..].as_mut_ptr(), tail_len);
+            }
+        }
+    }
+
+    #[inline]
+    pub fn map<T, F, const C: usize>(&self, r: DCacheRef, f: F) -> Result<T, DCacheError>
```
louisponet: I think we should have some comment that the reader needs to be able to "undo" whatever is done in `f` in case the data was stale/overwritten. I think also here we should not use self.reserved rather than something that's local in the data to determine whether we got sped past

crates/flux-utils/src/dcache.rs
```diff
@@ -0,0 +1,377 @@
...
+            return Err(DCacheError::StaleData);
+        }
+
+        Ok(result)
+    }
+
+    #[inline]
+    pub fn write<F>(&self, len: usize, f: F) -> Result<DCacheRef, DCacheError>
+    where
+        F: FnOnce(&mut [u8], &mut [u8]),
```
louisponet: Is this because of wrappping?

ninaiiad: yes

crates/flux-utils/src/dcache.rs
```diff
@@ -0,0 +1,413 @@
...
+/// synchronisation to the caller.
+///
+/// Intended use: the caller exchanges [`DCacheRef`]s through an external
+/// ordered channel (e.g. a spine queue) and the channel's Release/Acquire
+/// pair is the synchronisation point for the payload bytes stored here.
+#[repr(C)]
+pub struct DCache<const N: usize> {
+    reserved: AtomicUsize,
+    _pad: [u8; CACHELINE - size_of::<AtomicUsize>()],
+    data: Box<UnsafeCell<Aligned<N>>>,
```
louisponet: So one thing I'd do is have this just be a dynamic sized but inline [ ], so then we could in theory put this in shmem as well.

crates/flux-utils/benches/dcache.rs
```diff
@@ -0,0 +1,403 @@
...
+    ds_ix: usize,
+    r: DCacheRef,
+}
+
+const CAPACITY: usize = 64 * 1024 * 1024; // ~164 x 400KB messages before ring wraps
+const QUEUE_LEN: usize = 1024 * 16;
+const MSG_SIZE: usize = 409600; // 400kb
+
+#[inline]
+fn now_nanos() -> u64 {
```
louisponet, line 25: for these kind of benchmarks it's actually better to use Instant::now(), systemtime::now is 24ns each

crates/flux-utils/benches/dcache.rs
```diff
@@ -0,0 +1,403 @@
...
+
+    let producers: Vec<_> = (0..n_producers)
+        .map(|_| {
+            let tx = tx.clone();
+            thread::spawn(move || {
+                let mut rng = now_nanos();
+                for _ in 0..per_producer {
+                    let mut msg = vec![0u8; msg_size].into_boxed_slice();
+                    write_ts(&mut msg);
+                    tx.send(black_box(msg.clone())).unwrap();
```
louisponet, line 321: do we have to clone here?

crates/flux-utils/benches/dcache.rs
```diff
@@ -0,0 +1,403 @@
...
+                    thread::sleep(Duration::from_micros(xorshift(&mut rng) % 100));
```
louisponet: Also rather than sleep, you can do something like 
```
let curt = Instant::now() // the flux one btw;
let dur = Duration::from_micros(...);
while curt.elapsed() < dur {};
```

crates/flux-utils/benches/dcache.rs
```diff
@@ -0,0 +1,403 @@
...
+        }
+        sum
+    });
+
+    thread::sleep(Duration::from_millis(100));
+
+    let producers: Vec<_> = (0..n_producers)
+        .map(|_| {
+            let tx = tx.clone();
+            thread::spawn(move || {
```
louisponet, line 316: also make sure you use core_affinity:: pinning to different cores (use the last phys cores you have,  15 is the last one)

crates/flux-utils/src/dcache.rs
```diff
@@ -0,0 +1,413 @@
...
+    #[inline]
+    pub fn read(&self, r: DCacheRef, buf: &mut [u8]) -> Result<(), DCacheError> {
+        if r.len > N - Self::OFFSET {
+            return Err(DCacheError::DataLenExceedsCapacity(r.len, N - Self::OFFSET));
+        }
+        if r.len > buf.len() {
+            return Err(DCacheError::BufferTooSmall(buf.len(), r.len));
+        }
+
+        let base = self.data.get().cast::<u8>();
```
louisponet: So this stuff, did you check whether simply using an index directly might be faster?

crates/flux-utils/benches/dcache.rs
ninaiiad: otherwise the data is not copied so it's unfair to the dcache
the dcache however does not allocate and this does

louisponet: I think the vec![] into boxed slice already essentially does kinda the same thing doesn't it? But I see the point, I think it's probably fine like this then for now.

crates/flux-utils/src/dcache.rs
```diff
@@ -0,0 +1,380 @@
...
+        if !Self::offsets_match(base, offset_ix, r.offset) {
+            return Err(DCacheError::StaleData);
+        }
+
+        unsafe {
+            std::ptr::copy_nonoverlapping(
+                base.add(offset_ix + Self::OFFSET),
+                buf.as_mut_ptr(),
+                r.len,
+            );
```
vladimir-ea, line 70: what prevents the slice of the buffer that is being read from being written concurrently?

ninaiiad: nothing, the reader will find out whether or not this happened after the function has completed

ltitanb (approved):
lgtm

PR #42, opened 2026-03-11
feat(flux-ctl): consumer groups panel + TUI UX improvements

louisponet:
### Consumer Groups Panel
- `QueueHeader::active_groups()` reads non-empty group_labels + group_cursors slots
- Garbage guard: bails early if `label.len() > GROUP_LABEL_LEN` (prevents crash on old-format queues)
- `ConsumerGroupInfo` struct + `read_consumer_groups(flink)` for integration tests
- `ShmemCache::consumer_groups()` for zero-syscall reads in the TUI detail view
- `render_consumer_groups()` with Group/Cursor/Lag/Backlog columns, color-coded lag, visual bar
- Panel only appears for queues with active consumer groups

### TUI UX Improvements
- **Groups start collapsed** — reduces noise on startup; expand/collapse state preserved across refreshes
- **Hide dead segments** (default: on) — toggle with `a` key; dead segments are noise most of the time
- **Sort by activity** — new `SortMode::Activity` sorts by msgs/sec descending, using a 10-second rolling window for smoothed rates

### Tests
- Unit test: `active_groups_round_trip`
- TUI tests: consumer groups visibility, hide-dead, collapsed-default, activity sort cycle
- Smoke tests: `consumer_groups_readable_from_shmem`, `consumer_groups_empty_for_missing_or_data_segment`
- Live demo: slow consumer example in `live.rs`

### DIFF
--- a/crates/flux-communication/src/queue/mod.rs
+++ b/crates/flux-communication/src/queue/mod.rs
@@ -101,7 +101,7 @@ impl QueueHeader {
 
         // TODO: use better fix of potential data race, either spinlock or something
         // similar to what is done in the collaborative consume
-        std::thread::sleep(Duration::from_micros(rand::thread_rng().gen_range(0..10)));
+        std::thread::sleep(Duration::from_micros(rand::rng().random_range(0..10)));
 
         for i in 0..MAX_GROUPS {
             if self.group_labels[i] == key {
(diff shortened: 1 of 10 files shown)


ltitanb (approved):
lgtm

PR #59, opened 2026-03-24
feat(queue): add Queue::try_open_shared returning Result instead of panicking

louisponet:
## Problem

`Queue::open_shared()` calls `.expect()` on the result of `InnerQueue::open_shared()`, which panics if the queue is invalid (e.g. poisoned elements from a crashed writer, wrong element size). This makes it impossible for consumers to handle invalid queues gracefully.

The overseer hits this on startup when stale queues exist from a crashed builder:
```
thread 'main' panicked at queue/mod.rs:515:18:
Couldn't open shared queue, was it initialized?: ElementPoisoned(70)
```

## Solution

Add `Queue::try_open_shared()` — same logic as `open_shared()` but returns `Result<Self, QueueError>` instead of panicking. This lets consumers (like the overseer) skip or retry poisoned queues gracefully.

`open_shared()` is left unchanged for backward compatibility.

## Changes
- `crates/flux-communication/src/queue/mod.rs` — 9 lines added

---

<a href="https://app.devin.ai/review/gattaca-com/flux/pull/59" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/crates/flux-communication/src/queue/mod.rs
+++ b/crates/flux-communication/src/queue/mod.rs
@@ -516,6 +516,15 @@ impl<T: Copy> Queue<T> {
         }
     }
 
+    /// Open an existing shared-memory queue, returning an error instead of
+    /// panicking if the queue is invalid (e.g. poisoned, wrong element size).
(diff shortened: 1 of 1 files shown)


ltitanb (approved):
lgtm

PR #67, opened 2026-03-31
fix(tcp): auto-grow epoll event buffer to prevent connection starvation

louisponet:
## Problem

`Events::with_capacity(128)` in `TcpConnector` caused epoll event starvation when >128 connections were active (e.g. the data gather receiver).

Mio uses edge-triggered epoll — connections that don't fit in the 128-event batch are deferred to the next `poll()` cycle. With heavy connections dominating each batch, lighter connections experienced ~1s read delays. The kernel measured this as `rcv_rtt: 999ms`, auto-tuned `rcv_space` down to ~14KB, and advertised a tiny TCP receive window — causing the sender to be `rwnd_limited: 99.7%` of the time.

## Diagnosis (from `ss -tnip` on both sides)

**Sender side:**
- `rwnd_limited: 99.7%` — blocked on receiver's window almost all the time
- `snd_wnd: 73728` (72KB) — tiny window from receiver
- `Send-Q: 931816` — ~1MB queued in kernel

**Receiver side (starved connection):**
- `rcv_space: 14480` (14KB) — kernel auto-tuned window down
- `rcv_rtt: 999ms` — kernel thinks app takes ~1s between reads

**Receiver side (healthy connection):**
- `rcv_space: 4734960` (4.7MB) — normal
- `bytes_received: 3.8GB` — orders of magnitude more throughput

## Fix

- After each `poll()`, if the number of returned events equals the buffer capacity, double it
- Log at `info` level when growth occurs (`"epoll event buffer full, growing"`)
- Add `.with_event_capacity(n)` builder method for callers that want a different starting size
- Default remains 128 — auto-grows as needed with zero overhead once stabilized

---

<a href="https://app.devin.ai/review/gattaca-com/flux/pull/67" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/crates/flux-network/src/tcp/connector.rs
+++ b/crates/flux-network/src/tcp/connector.rs
@@ -4,7 +4,7 @@ use flux::spine::{SpineProducerWithDCache, SpineProducers};
 use flux_timing::{Duration, Instant, Nanos, Repeater};
 use flux_utils::{DCachePtr, safe_panic};
 use mio::{Events, Interest, Poll, Token, event::Event, net::TcpListener};
-use tracing::{debug, error, warn};
+use tracing::{debug, error, info, warn};
 
 use crate::tcp::{
     ConnState, TcpStream, TcpTelemetry,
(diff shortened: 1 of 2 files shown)


gd-0:
no activity on this for a while so closing but keeping the branch alive

PR #82, opened 2026-05-21
add Timer::record_scoped RAII guard

bartoszmodelski:
## Summary
- New `Timer::record_scoped()` method that starts the timer and returns a `ScopedRecord` guard
- Guard calls `record_processing()` automatically on drop

Usage:
```rust
fn my_function(&mut self) {
    let _guard = self.timer.record_scoped();
    // ... work ...
} // record_processing() fires here
```

## Test plan
- [x] `cargo check -p flux-communication`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/crates/flux-communication/src/timer.rs
+++ b/crates/flux-communication/src/timer.rs
@@ -144,6 +144,7 @@ impl Timer {
     }
 
     #[inline]
+    #[must_use = "the returned guard records on drop; binding it to `_` drops it immediately"]
     pub fn record_scoped(&mut self) -> ScopedRecord<'_> {
         self.start();
         ScopedRecord { timer: self }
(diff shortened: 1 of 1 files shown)


gd-0 (approved):

crates/flux-communication/src/timer.rs
```diff
@@ -143,6 +143,12 @@ impl Timer {
         self.emit_processing();
     }
 
+    #[inline]
+    pub fn record_scoped(&mut self) -> ScopedRecord<'_> {
```
niki4smirn, line 148: `#[must_use]`

just in case :)

niki4smirn (approved):

PR #93, opened 2026-06-25
experimental feature: work on signal

vladimir-ea:
adds a `park` feature to Flux. 

If this feature is enabled threads will be parked if no work was done on a loop and woken whenever a message is added to a spine queue. 

Also supports network triggers via `mio::Waker`. 

On linux the thread parking / waking uses futex.

### DIFF
--- a/crates/flux-network/src/tcp/stream.rs
+++ b/crates/flux-network/src/tcp/stream.rs
@@ -700,6 +700,7 @@ impl TcpStream {
 /// Set `TCP_USER_TIMEOUT` on a mio `TcpStream`.
 /// After this duration of unacknowledged data the kernel closes the connection,
 /// overriding the system-wide `tcp_retries2` (~15 min default) for this socket.
+#[cfg(target_os = "linux")]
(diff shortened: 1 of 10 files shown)


0w3n-d (approved):

0w3n-d:
Looks good to me. The only issue for me is that some tiles that are not busy will be woken up by unrelated activity in other tiles due the global nature of how it works. It could lead to a lot of park/wake churn. But I think it's fine for now. The extra context needed to avoid that isn't worth the added complexity.

vladimir-ea:
> Looks good to me. The only issue for me is that some tiles that are not busy will be woken up by unrelated activity in other tiles due the global nature of how it works. It could lead to a lot of park/wake churn. But I think it's fine for now. The extra context needed to avoid that isn't worth the added complexity.

I did consider how this might be done. The 32-bit counter passed to futex is used to track whether any new work was added whilst threads are working. Some bits might be stolen to signal which queues has work added - or a separate set of fields might be used - but it's not obvious when those should be cleared or reset. 

Running the current impl, cores for tiles that do very little report 0% cpu use - so although they are woken constantly it appears to be cheap at least in cpu terms.

PR #96, opened 2026-07-07
Flux profiler

alexX512:


crates/flux-communication/src/queue/mod.rs
```diff
@@ -506,9 +506,12 @@ impl<T: Copy> InnerQueue<T> {
         }
         let shmem = ShmemConf::new().flink(path).open()?;
         let ptr = shmem.as_ptr();
-        std::mem::forget(shmem);
         // len=0: skip floor check, still validates elsize + poisoned slots.
-        Self::open_initialized(ptr, 0)
+        let opened = Self::open_initialized(ptr, 0);
+        if opened.is_ok() {
+            std::mem::forget(shmem);
```
louisponet, line 512: why only if is_ok?

crates/flux-profiler-macros/src/lib.rs
```diff
@@ -0,0 +1,60 @@
...
+/// (no hot-path cost); the report unwraps the marker into a `fn<Type>` label.
+/// Free functions keep the plain `module::path::fn` name.
+///
+/// Pass a string literal to override: `#[timed("custom_name")]` uses that name
+/// verbatim (no module prefix, no type qualification).
+///
+/// Records the frame's close on every exit path — normal return, `?`, early
+/// `return`, panic-unwind — via a Drop guard.
+#[proc_macro_attribute]
+pub fn timed(attr: TokenStream, item: TokenStream) -> TokenStream {
```
louisponet, line 35: can we have a feature flag to enable disable this?

crates/flux-profiler/src/perf/events.rs
```diff
@@ -0,0 +1,248 @@
...
+    },
+    Metric {
+        name: "l3-misses",
+        spec: Spec::PerVendor { amd: Enc::Raw(0xfc43), intel: Enc::Cache(2, 0, 1) },
+    },
+];
+
+/// Resolve a curated metric for the running CPU.
+fn canonical(name: &str) -> Option<EventSpec> {
+    let m = METRICS.iter().find(|m| m.name == name)?;
```
louisponet, line 135: Given that it's a const fixed size array, wouldn't an actual fixed size array + a enum that can be used to directly index into it be faster?

crates/flux-profiler/src/perf/source.rs
```diff
@@ -0,0 +1,41 @@
...
+    fn open() -> Option<Self> {
+        let opened: Vec<_> =
+            Schema::local().iter().map(|e| HwCounter::event(e.type_, e.config)).collect();
+        opened.iter().any(Option::is_some).then_some(Self { opened })
+    }
+
+    #[inline]
+    fn read(&self) -> PerfSample {
+        let mut s = PerfSample::default();
+        for (i, c) in self.opened.iter().enumerate().take(MAX_EVENTS) {
```
louisponet: I don't quite get this. Why open all but then only ever take max events? Aren't we always taking 1 event per opened?

crates/flux-profiler/src/fxt.rs
```diff
@@ -0,0 +1,339 @@
...
+    threads: impl Iterator<Item = ThreadEvents<'a>>,
+    meta: &FlamegraphMeta,
+) -> Vec<u8> {
+    let FlamegraphMeta { names, schema } = meta;
+    let mut threads: Vec<_> = threads.collect();
+    threads.sort_by(|a, b| a.name.cmp(b.name).then(a.tid.cmp(&b.tid)));
+    debug_assert!(threads.len() < 256, "thread ref is 8-bit");
+    let now_tsc = Instant::now().0 & TSC_MASK;
+    let now_epoch_ns =
+        SystemTime::now().duration_since(UNIX_EPOCH).map_or(0, |d| d.as_nanos() as u64);
```
louisponet: ?

crates/flux-profiler/src/ring_drainer.rs
```diff
@@ -0,0 +1,178 @@
...
+        // Collaborative so the cursor starts at the beginning of the ring, `next_seq`
+        // at 0
+        consumer.try_init_collaborative();
+        Ok(Self { consumer, pending: VecDeque::new(), next_seq: 0, missed: 0 })
+    }
+
+    fn drain(&mut self) {
+        let mut scratch = T::default();
+        loop {
+            match self.consumer.try_consume(&mut scratch) {
```
louisponet, line 38: I think this can become a spinloop never exiting, wonder if it's not best to have some timeout

crates/flux-profiler/src/ring_drainer.rs
```diff
@@ -0,0 +1,178 @@
...
+                Ok(()) => {
+                    self.pending.push_back(scratch);
```
louisponet, line 40: Not sure if it's always sampling but you probably can do something similar where you only track a certain amount of samples like in the timekeeper

crates/flux-profiler/README.md
```diff
@@ -0,0 +1,121 @@
+# flux-profiler
+
+A cross-process flamegraph profiler for flux. Annotate functions with `#[timed]`,
```
bartoszmodelski: tracing profiler

crates/flux-profiler-macros/src/lib.rs
bartoszmodelski: +1, a global flag to compile with or without timers (can rely on compiler remover empty functions) 

if it's not more difficult, I think ideally we could tag the timers and enable only the ones we care about: simulator timers, greedy timers, etc. but maybe that's for later

bartoszmodelski: +1, a global flag to compile with or without timers (can rely on compiler remover empty functions) 

if it's not more difficult, I think ideally we could tag the timers and enable only the ones we care about: simulator timers, greedy timers, etc. but maybe

crates/flux-profiler/src/perf/events.rs
```diff
@@ -0,0 +1,248 @@
...
+    Metric { name: "l1i-misses", spec: Spec::Any(Enc::Cache(1, 0, 1)) },
+    Metric {
+        name: "l2-misses",
+        spec: Spec::PerVendor { amd: Enc::Raw(0xfe43), intel: Enc::Raw(0x10d1) },
+    },
+    Metric {
+        name: "l3-misses",
+        spec: Spec::PerVendor { amd: Enc::Raw(0xfc43), intel: Enc::Cache(2, 0, 1) },
+    },
+];
```
bartoszmodelski, line 131: Is there any way to resolve these on demand?  There's prob >100 counters

crates/flux/src/park.rs
```diff
@@ -56,6 +58,12 @@ pub struct Signal {
 
 pub static SIGNAL: Signal = Signal::new();
 
+impl Default for Signal {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
```
bartoszmodelski, line 66: is this needed?

crates/flux-profiler/src/symbols.rs
```diff
@@ -0,0 +1,100 @@
...
+    path: PathBuf,
+}
+
+impl CrossProcessSymbolsResolver {
+    pub(super) fn new(pid: u32) -> Self {
+        Self { segments: parse_maps(pid) }
+    }
+}
+
+impl FrameResolver for CrossProcessSymbolsResolver {
```
bartoszmodelski, line 50: is there any llvm library that could do this for us?

crates/flux-profiler/src/timing.rs
```diff
@@ -0,0 +1,58 @@
...
+
+/// Drop-based timer scope used by the `#[timed]` macro expansion. Records a
+/// frame open on construction and a close on every exit path — normal return,
+/// `?`, early `return`, panic-unwind.
+#[doc(hidden)]
+pub struct TimerGuard {
+    close: Option<&'static str>,
+}
+
+impl TimerGuard {
```
bartoszmodelski, line 17: is this what #[timer] macro compiles into?

crates/flux/src/park.rs
vladimir-ea: clippy says yes :-P

crates/flux-communication/src/queue/mod.rs
alexX512: mem::forget keeps the mapping alive for the returned ptr. On error nothing uses it, so we let it drop/unmap, I had one issue with descriptors leak because of that

crates/flux-profiler/src/timing.rs
alexX512: Yes, it creates defer with some codegen to have function names in .rodata section

crates/flux-profiler/src/perf/events.rs
alexX512: There is some C library which has much wider counters support, didn't found smth similar for Rust, so for now just manual implementation with small subset of counters

Potentially can try to integrate C lib in the future

crates/flux-profiler/src/symbols.rs
alexX512: Crates cover the pieces (procfs/proc-maps, object) but none does the full cross-process virtual address -> file-offset -> read-bytes from .rodata section of the file. Only the ~10-line maps parse overlaps, so not really worth a dep

crates/flux-profiler/src/perf/events.rs
alexX512: This one is just parse perf schema from disk which producer dump on init, so not really performance oriented and it is just easier to do lookup by strings from a file

crates/flux-profiler/src/perf/source.rs
alexX512: We take up to MAX_EVENTS from env var as there is no more slots in the perf message, will limit it in constructor to not have that check in read

crates/flux-profiler/src/fxt.rs
alexX512: True that IngestionTime can be used here, will switch

crates/flux-profiler/src/ring_drainer.rs
alexX512: Read should be much faster than write as we usually do function work between events, can add batch reading up to X elements per drain just in case

crates/flux-profiler/src/ring_drainer.rs
alexX512: It is intentionally instrumented and not sampled version of profiler so we can correctly rebuild all stack traces, with that we need all events to not have corrupted stacks

crates/flux-profiler-macros/src/lib.rs
alexX512: Added `disable-profiling` flag, so by default if someone just want to check that profiler works there will be no need to think about flags. But if we want to have fully zero overhead build we can disable it

bartoszmodelski (approved):

crates/flux-profiler-macros/src/lib.rs
louisponet: I like that!

PR #105, opened 2026-07-22
set niceness rather than prio

bartoszmodelski:
Fifo is footguny: on one hand we can starve others, on the other os by default reserves some % of time to non-fifo tasks so we may occasionally get starved as well. Let's instead just set the niceness to max and isolate the cores well.


crates/flux-utils/src/thread.rs
```diff
@@ -65,10 +41,22 @@ pub fn get_tid() -> i64 {
     0
 }
 
-pub fn thread_boot(core: Option<usize>, prio: ThreadPriority) {
+pub fn thread_boot(core: Option<usize>, niceness: Option<i32>) {
```
gd-0: contain niceness in the emum we had before

crates/flux-utils/src/thread.rs
```diff
@@ -65,10 +41,22 @@ pub fn get_tid() -> i64 {
...
+mod tests {
+    use super::*;
+
+    #[cfg(target_os = "linux")]
+    #[test]
+    #[should_panic(expected = "thread niceness must be between -20 and 19")]
+    fn rejects_invalid_niceness() {
+        validate_thread_niceness(20);
+    }
 }
```
gd-0: bs llm test :D

gd-0 (changes requested):

crates/flux-utils/src/thread.rs
bartoszmodelski: actually makes even more sense now since num is inversely proportional to the priority - amended

gd-0 (approved):

PR #122, opened 2026-08-12
feat(network): export FRAME_HEADER_SIZE and write_frame_header

Sham-BahamasTrading:
gattaca-com/builder#2872 produces flux-framed TCP broadcasts from the RPC's serialisation ring, which `TcpConnector` cannot consume (multi-producer, and the same ring also feeds the legacy byte-stream port). It currently restates the 12-byte `[u32 len LE][u64 send_ts LE]` header locally; review asked for it to come from flux instead.

This makes `FRAME_HEADER_SIZE` and `write_frame_header` public and re-exports them from `tcp`. No behaviour change.

Branched off `ce40087` (the rev builder currently pins) so builder can bump straight to this commit without pulling anything else in.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/crates/flux-network/src/tcp/mod.rs
+++ b/crates/flux-network/src/tcp/mod.rs
@@ -2,4 +2,4 @@ mod connector;
 mod stream;
 
 pub use connector::{PollEvent, SendBehavior, TcpConnector};
-pub use stream::{ConnState, TcpStream, TcpTelemetry};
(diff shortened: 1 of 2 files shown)


Sham-BahamasTrading:
Closing — the plan changed to a full-flux send side in gattaca-com/builder#2872 (orders go through `TcpConnector::write_or_enqueue_with`, which frames internally), so nothing external needs the header writer after all.

PR #128, opened 2026-08-17
feat: versioned types

jayrabjohns:


crates/flux-versioned-types-macros/src/evolve_enum/mod.rs
```diff
@@ -51,5 +51,12 @@ use syn::parse_macro_input;
 pub fn evolve_enum(input: TokenStream) -> TokenStream {
     let mut input = parse_macro_input!(input as EvolveEnumInput);
     input.ensure_default_attrs(crate::shared::default_enum_attrs);
-    crate::shared::generate_evolving(&mut input, generate_base_enum, generate_evolution).into()
+    let require_type_hash_locks = input.require_type_hash_locks;
```
alexX512: Might be we can always say that require_type_hash_locks is true? As I don't think we want to use evolve enums without locking their hash

crates/flux-versioned-types-macros/src/evolve/mod.rs
```diff
@@ -0,0 +1,109 @@
...
+/// ```
+///
+/// This generates:
+/// - `MyTypeV1 { a: u32, b }` with `default_attrs` + `type_hash_lock`
+/// - `MyTypeV2 { a: u32, b, c }` with `Into<MyTypeV2> for MyTypeV1`
+/// - `MyTypeV3 { a: u64, c, d }` with `Into<MyTypeV3> for MyTypeV2`
+pub fn evolve_struct(input: TokenStream) -> TokenStream {
+    let mut input = parse_macro_input!(input as EvolveInput);
+    input.ensure_default_attrs(crate::shared::default_struct_attrs);
+    let require_type_hash_locks = input.require_type_hash_locks;
```
alexX512: Same as above comment

crates/flux-versioned-types/scripts/add-missing-typehashes.sh
```diff
@@ -0,0 +1,124 @@
+#!/usr/bin/env bash
```
alexX512, line 1: Btw how that script can be used when flux is just a crate dependency in a different repo? Might be we can add some type of readme

alexX512 (approved):

PR #135, opened 2026-08-25
Flux-network re-design

Bronek:
Includes documents:

- CONTEXT.md: the workspace glossary (Tile, Spine, Signal, Waker; Stream, Group, Tenant, Raw group, Owned/External mode, Endpoint, Peer, Deadline, Draining, Lingering, Refused).
- ADR 0001: poll ownership is a construction-time mode of the stream network, and protocol layers are tenants the network schedules — one driver owns deadline folding, routing by group, maintenance order and ticking; tenants expose their events by pull; hooks stay private behind an opaque TenantRef carrier.
- ADR 0002: transports are the closed set Endpoint { Tcp, Unix }.
- ADR 0003: Unix-domain socket files are probed (lstat, then connect) before a stale one is replaced, and removed on close.
- docs/extension-points.md: eventual (streamed responses) and speculative (TLS, Unix socket file mode, QUIC) directions the design keeps open without deciding.

Refer: #143


CONTEXT.md
```diff
@@ -0,0 +1,90 @@
...
+**Stream**:
+An ordered byte channel with one peer at each end — a TCP or Unix-domain connection — and the
+unit a stream network drives. A response held open for appended writes is not a stream in this
+sense.
+_Avoid_: socket (the OS handle, not the channel), pipe
+
+**Group**:
+A set of connections sharing one wire framing and one socket configuration inside a
+network, with limits enforced per connection.
+_Avoid_: pool, channel, protocol
```
vladimir-ea: not sure i understand what this is in practice - would a Http Server listening on a socket address constitute a 'group'?

CONTEXT.md
```diff
@@ -0,0 +1,90 @@
...
+
+**Tenant**:
+A protocol layer that owns exactly one Group inside a shared network and is scheduled by
+that network; an HTTP server or client is a tenant.
+_Avoid_: instance, service, sub-network
```
vladimir-ea: seems like Tenant and Group describe the same thing? what is the difference? These are not commonly used terms - there must be existing terminology for these things and if there isn't maybe they are the wrong categories (this is well-trodden territory)

CONTEXT.md
```diff
@@ -0,0 +1,90 @@
...
+_Avoid_: unmanaged group, bare group
+
+**Owned mode**:
+A network that holds its own poll and drives it, with whatever timeout the caller passes.
+_Avoid_: standalone, embedded
+
+**External mode**:
+A network built over a poll the caller holds; the caller delivers readiness events and
+drives timers, and may register its own sources alongside.
+_Avoid_: injected, shared, hosted
```
vladimir-ea, line 63: if these relate to `Poll` let's call them `OwnedPoll` and `ExternalPoll`

CONTEXT.md
vladimir-ea: The first ADR literal says 'tenants are protocols1 - can we use the term 'protocol'?

CONTEXT.md
```diff
@@ -0,0 +1,90 @@
...
+
+**Tenant**:
+A protocol layer that owns exactly one Group inside a shared network and is scheduled by
+that network; an HTTP server or client is a tenant.
+_Avoid_: instance, service, sub-network
+
+**Raw group**:
+A Group no Tenant owns; the caller consumes its events directly as they arrive, borrowing
+each payload for the duration of the call.
+_Avoid_: unmanaged group, bare group
```
vladimir-ea: what does this look like? if a tenant is a protocol what is this? a raw byte stream a tcp connection? could we call it a 'byte stream' or 'tcp connection'?

CONTEXT.md
Bronek: A group is how we manage connections (incoming and outgoing) into one bucket. There can be arbitrary number of groups (e.g. HTTP server to exposed to internal network **only** can be less strict in enforcing timeouts on connections and have its own connection limit and max frame size), but all groups will share the same Tile. A group can hold both client and server ("tenant"). This not a new concept, there already was `TcpGroup` and `TcpGroupConfig` #133 but we are generalizing it.

CONTEXT.md
Bronek: Tenant is how we schedule work on the server side - Group also holds outbound connections. A tenant holds `HttpConfig` i.e. the part of configuration that applies not to a connection as such, but specifically to serving HTTP requests - both over TCP and Unix-domain sockets. Considered "protocol" but the term is heavily overloaded, it would likely cause confusion. More suitable alternatives `Layer` `Handler` `Service` `Adapter` - but `Tenant` seem least ambiguous. Group owns connections and Tenant owns (protocol) work over these connections. There is also notion of "Raw group" (an unclaimed Group) - a set of connections sharing one configuration, which have no Tenant (example, if we wanted to support these: gRPC, WebSocket; or a dumb protocol e.g. a proxy; or a persistent outbound feed; or flux's own wire protocol #107 )

CONTEXT.md
Bronek: No need to create publicly accessible names for these - `enum PollMode { Owned, External }` is private, the public surface is for the user to call either:
- `StreamNetwork::default() ` , Tile will be internally created and owned by flux, or
- `StreamNetwork::with_registry(registry, token_base)` - Tile must be created by user, with a user-coded loop to poll and handle events

CONTEXT.md
Bronek: It's neither a connection (group can have many of these) not raw bytes. Will rename to "Unclaimed group"

CONTEXT.md
vladimir-ea: i think 'service' would be better understood

CONTEXT.md
vladimir-ea: so the http server in your example would be a group? i still struggle to understand what it is

CONTEXT.md
vladimir-ea: ok lets at least put the word Poll in the title of these doc sections then

CONTEXT.md
vladimir-ea: i still don't understand what a group is ... maybe its me, but this terminology is confusing

CONTEXT.md
Bronek: an HTTP server, or a client, requires both a Group (how connections are managed) and a Tenant (what happens over *these* connections)

```Rust
let group  = net.add_group(StreamGroupConfig { framing: Framing::Raw, max_connections: Some(256), .. }); // the bucket: sockets + transport policy
let server = HttpTenant::new(&mut net, group, HttpConfig { idle_timeout: .., .. });                 // the protocol on that bucket
server.listen(&mut net, Endpoint::Tcp(addr))?;     // sockets go into the bucket
server.listen(&mut net, Endpoint::Unix(path))?;
```

Bronek: Renamed `Group` to `ConnectionGroup` in 4c5381837e1d515455c00d061954cfab4d7ec1d9

Could not use Transport - too heavily used in context where it implies single underlying protocol, whereas single `ConnectionGroup` can own both TCP and Unix-domain sockets

CONTEXT.md
Bronek: Renamed to Service in 4c5381837e1d515455c00d061954cfab4d7ec1d9

CONTEXT.md
Bronek: Changed to `Owed poll` / `External poll` in 4c5381837e1d515455c00d061954cfab4d7ec1d9

CONTEXT.md
Bronek: Changed to `Unclaimed ConnectionGroup` in 4c5381837e1d515455c00d061954cfab4d7ec1d9

CONTEXT.md
vladimir-ea: maybe a box diagram would be useful to show the relationships of the differnet things?

Bronek:
Naming rationale for `ConnectionGroup` and `Service`:

---

# Why `ConnectionGroup` and `Service`

## Decision

Use `ConnectionGroup` for the network-owned grouping abstraction and `Service` for the application-layer server or client scheduled over it.

The core ownership rule is:

```text
StreamNetwork
└── ConnectionGroup
    ├── shared connection policy
    ├── TCP and Unix-domain connection machinery
    └── one protocol owner
        └── Service, such as HttpService
            (or the caller, for an unclaimed group)
```

A `Service` owns exactly one `ConnectionGroup`. `HttpService` is the first implementation. One instance can act as HTTP server, HTTP client, or both at once. A service is passed to the network as a `ServiceRef`, produced by `as_service()`; that carrier is an API detail rather than a separate domain concept.

## Terms

### `ConnectionGroup`

The inbound and outbound connections—TCP or Unix-domain—that share one configuration and one protocol owner. Its configuration covers framing, socket policy, queues or backlog, connection caps, and related resource limits.

The group also owns listeners and persistent outbound connections that reconnect on their own. The API calls the latter *outbound endpoints*. In the glossary sense an endpoint is only an address: what a listener binds or an outbound connection dials.

### `Service`

An application-layer server or client. It owns one `ConnectionGroup` and is scheduled by its network. Currently this is implemented by `HttpService`, for both HTTP server and client roles.

ADR wording can remain general: services are protocol layers scheduled by the network; `HttpService` is the first implementation. **Hypothetical** other services living in this same abstraction layer: `GrpcService` , `WebSocketService`.

### Unclaimed `ConnectionGroup`

A `ConnectionGroup` that no `Service` has claimed. The caller is its protocol layer and receives its events directly while driving the network. Claiming is a whole-group decision made when a `Service` is constructed and reversed when it closes; it is never per-connection state.

## Why not `Tenant` or `Protocol`?

`Tenant` suggests SaaS isolation or multi-tenancy, neither of which describes the abstraction. It also over-generalises from a design whose only production implementation is HTTP.

`Protocol` is heavily overloaded in Rust and normally names a protocol kind, such as HTTP, rather than a live application-layer instance doing work. A service *implements* a protocol role; it is not itself the protocol.

`Service` describes the active instance and fits both the HTTP client and server roles without implying a particular wire protocol.

## Why not call the group a `Transport`?

TCP and Unix-domain sockets are transport kinds. The grouping abstraction is not one transport: it may contain connections of both kinds while imposing one configuration and ownership boundary. Calling that heterogeneous scope a `Transport`, `Transports`, or `TransportGroup` would blur the distinction between the underlying machinery and the policy scope that contains it. The crate already uses *transport* for exactly that kind distinction: `TransportStream`, `transport.rs`, and the `over_both_transports!` test macro. "A Transport mixing TCP and Unix listeners" would then be a legal sentence.

`StreamGroup` was precise, but `ConnectionGroup` uses more familiar networking vocabulary and better signals that the abstraction owns connection lifecycle, not merely byte streams.

Other candidates were weaker:

- `Lane` is a coined metaphor without established networking meaning.
- `Pool` suggests interchangeable reusable resources.
- `Channel`, `Flow`, and `Session` imply different lifetimes or semantics.
- `Endpoint` already means an address in this model.
- `Stack` is broader than the ownership and scheduling scope.

## The homogeneous-`Transport` alternative

An attractive alternative was to define separate homogeneous `TcpTransport` and `UnixTransport` groupings, each with configuration matching its underlying machinery, and require every `Service` to own exactly one of them.

That model is clean only while a service uses one transport kind. A logical HTTP service supporting both TCP and Unix-domain sockets would force one of four outcomes:

1. Two service instances of one implementation. Each has its own state   machine, readiness queue, deadlines, counters, and close call. The caller   dispatches events from both and must track which service a token belongs   to before answering by token.
2. A composite service that merges two child services, thereby introducing a   second scheduler for work, deadlines, fairness, telemetry, and responder   ownership.
3. One shared HTTP state machine owning two transports, which breaks the   one-service/one-transport rule and reintroduces multi-transport grouping.
4. A union or composite transport, which recreates `ConnectionGroup` under a   different name.

The homogeneous model therefore optimises the configuration type at the cost of leaking transport coordination into the application layer. Keeping one `Service`, one `ConnectionGroup`, and one connection lifecycle is simpler, more predictable, and easier to optimise. The direction stays recorded as speculative in `extension-points.md`, with its trigger: a service needing different policy per transport.

## Preserve transport-specific configuration explicitly

The useful part of the homogeneous-transport proposal is structural clarity for transport-specific options. Keep the heterogeneous `ConnectionGroup`, but nested TCP-only policy under `tcp`:

```rust
pub struct ConnectionGroupConfig {
    // ... fields shared by TCP and Unix-domain connections:
    // framing, caps, backlog, reconnect interval, telemetry ...
    /// TCP-only socket options; Unix-domain connections ignore them.
    pub tcp: TcpOptions,
}

pub struct TcpOptions {
    pub nodelay: bool,
    pub keepalive: bool,
    pub user_timeout_ms: u32,
}
```

`TcpOptions` should have defaults and need not be optional. An `Option<TcpOptions>` would be ambiguous: absence could mean either “use TCP defaults” or “TCP is forbidden.” Common policy—framing, buffer sizes where applicable to both socket kinds, reconnection, backlog and frame limits, connection caps, and telemetry—remains at group level. A `UnixOptions` type should appear only when genuine configurable Unix-specific policy exists (e.g. hypothetical file permissions).

## Architectural consequence

`ConnectionGroup` is the containment boundary that keeps connection kinds, lifecycle, resource policy, and ownership out of the service implementation. Removing that boundary does not remove the complexity; it makes every service coordinate transports itself.

The chosen vocabulary supports the intended engineering principles:

- **Performance:** one state machine, lifecycle, readiness path, and scheduler  per logical service.
- **Simplicity:** names describe familiar roles without pretending TCP and Unix  require separate application-layer instances.
- **Containment:** transport-specific mechanics remain inside the network and   group; the service owns one explicit application-layer scope.

CONTEXT.md
vladimir-ea: also not sold on 'transport' being singular - planes, trains and automobiles = transport. but fine.

vladimir-ea (approved):

CONTEXT.md
Bronek: diagram and explanation in https://github.com/gattaca-com/flux/pull/135#issuecomment-5425639638

PR #138, opened 2026-08-26
flux-network: support TCP and Unix-domain streams

Bronek:
This PR builds on #137 by adding Unix-domain sockets to `StreamNetwork`. TCP and Unix-domain listeners and outbound connections can coexist in one network and one `ConnectionGroup`.

## Transport model

`StreamNetwork::listen` and `connect` now take an explicit `Endpoint`:

```rust
pub enum Endpoint {
    Tcp(SocketAddr),
    Unix(PathBuf),
}
```

Events identify the remote end as `Peer::Tcp(SocketAddr)` or `Peer::Unix`. Unix clients are anonymous because they do not bind a path of their own.

The two transports share the same connection lifecycle, framing and queue machinery. Private enums dispatch to the corresponding mio listener or stream with one match per operation. This keeps mixed-transport groups possible without generic networks or hot-path virtual calls, as decided in ADR 0002.

`HttpNetwork` accepts the same endpoint set. For an outbound Unix connection, it writes `Host: localhost` when the caller supplies no `Host` header.

## Unix socket paths

Binding a Unix listener follows ADR 0003:

- A missing path is bound normally.
- A stale socket file is removed and replaced.
- A live listener, including one with a full accept queue, returns `AddrInUse`.
- A regular file, directory or symlink is left untouched and returns `AlreadyExists`.
- Other probe failures preserve the existing object and return an error naming its path.

The ownership probe is nonblocking. Dropping the listener removes the socket file it created; after an unclean exit, the next bind detects and replaces the stale file. Outbound Unix endpoints retry at the group's reconnect interval, like TCP endpoints.

Socket-file permissions come from the process umask. Flux does not set a mode or owner.

## Configuration

TCP-only policy is grouped explicitly:

```rust
pub struct ConnectionGroupConfig {
    pub tcp: TcpOptions,
    // Common connection policy...
}
```

`TcpOptions` contains `nodelay`, `keepalive` and `user_timeout_ms`, preserving their previous defaults. Unix connections have no corresponding options. Socket buffer sizing remains common to both transports.

`Endpoint` intentionally has no `From<SocketAddr>` implementation: callers choose the transport explicitly.

The default group name changes from `"tcp"` to `"stream"`. This changes the telemetry suffix for callers that use the default name; the existing `tcp_latency_*` and `tcp_alloc_*` metric prefixes do not change.

## Scope and verification

`listen` still returns `io::Result<()>`; returning the bound endpoint follows later in the stack. `HttpNetwork` remains callback-driven, and this PR adds no scheduling or connection-cap behavior.

The stream-network and HTTP suites exercise shared behavior over both transports. Dedicated lifecycle tests cover stale and live socket files, a saturated listener, non-socket filesystem objects, datagram sockets and unlink-on-close. A unit test pins the `TcpOptions` defaults. Workspace tests, Clippy with warnings denied, and the nightly formatting check pass at this tip.

## Known test flakes

The two pre-existing races described in #137 remain at this stack boundary. One waits a fixed five seconds for a backpressure marker and is fixed in A3; the other guesses an unused TCP port and is fixed later when `listen` can report the endpoint it bound. If either appears in CI, rerun the failed job.

Base: #137. Next: A3, adding network-scheduled services and `HttpService`.

Refer: #143

Assisted-by: Codex:gpt-5

### DIFF
--- a/crates/flux-network/src/stream/endpoint.rs
+++ b/crates/flux-network/src/stream/endpoint.rs
@@ -1,0 +1,59 @@
+use std::{fmt, net::SocketAddr, path::PathBuf};
+
+/// The address a listener binds or an outbound connection targets.
+///
+/// The set of transports is closed: a caller that accepts addresses as text
+/// parses them itself and builds the variant it wants.
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
+pub enum Endpoint {
+    /// A TCP socket address.
+    Tcp(SocketAddr),
+    /// A Unix-domain socket path.
+    Unix(PathBuf),
+}
+
+impl Endpoint {
+    /// The identity every connection to this endpoint reports.
+    pub(crate) fn peer(&self) -> Peer {
+        match self {
+            Self::Tcp(addr) => Peer::Tcp(*addr),
+            Self::Unix(_) => Peer::Unix,
+        }
+    }
+}
+
+impl From<SocketAddr> for Endpoint {
+    fn from(addr: SocketAddr) -> Self {
+        Self::Tcp(addr)
+    }
+}
+
+impl fmt::Display for Endpoint {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            Self::Tcp(addr) => write!(f, "{addr}"),
+            Self::Unix(path) => write!(f, "{}", path.display()),
+        }
+    }
+}
+
+/// The identity of the remote end of a connection.
+///
+/// A Unix-domain client is anonymous: it binds no path of its own, so the
+/// kernel reports an unnamed address for it.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum Peer {
+    /// The remote TCP socket address.
+    Tcp(SocketAddr),
+    /// The unnamed remote end of a Unix-domain connection.
+    Unix,
+}
+
+impl fmt::Display for Peer {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            Self::Tcp(addr) => write!(f, "{addr}"),
+            Self::Unix => f.write_str("unix"),
+        }
+    }
+}
--- a/crates/flux-network/src/stream/endpoint.rs
+++ b/crates/flux-network/src/stream/endpoint.rs
@@ -22,12 +22,6 @@ impl Endpoint {
     }
 }
 
-impl From<SocketAddr> for Endpoint {
-    fn from(addr: SocketAddr) -> Self {
-        Self::Tcp(addr)
-    }
-}
-
 impl fmt::Display for Endpoint {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         match self {
--- a/crates/flux-network/src/stream/network.rs
+++ b/crates/flux-network/src/stream/network.rs
@@ -1,20 +1,21 @@
 use std::{
     io::{self, IoSlice, Read, Write},
-    net::{Shutdown, SocketAddr},
+    net::Shutdown,
     ops::{Deref, DerefMut},
 };
 
 use flux_communication::Timer;
 use flux_timing::{Duration, Instant, Nanos, Repeater};
-use mio::{Events, Interest, Poll, Registry, Token, event::Event, net::TcpListener};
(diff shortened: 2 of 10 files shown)


crates/flux-network/src/stream/endpoint.rs
```diff
@@ -0,0 +1,53 @@
...
+/// The identity of the remote end of a connection.
+///
+/// A Unix-domain client is anonymous: it binds no path of its own, so the
+/// kernel reports an unnamed address for it.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum Peer {
+    /// The remote TCP socket address.
+    Tcp(SocketAddr),
+    /// The unnamed remote end of a Unix-domain connection.
+    Unix,
```
vladimir-ea, line 43: since this implements `Eq` and `PartialEq` all `Unix` peers will appear to be the same? why not include the path as the equivalent of the `SocketAddr`?

vladimir-ea: i see from the comment above that maybe this is an issue, but isn't having all be 'eq' also an issue?

crates/flux-network/src/stream/network.rs
```diff
@@ -339,35 +325,35 @@ impl NetworkState {
...
             return false;
         }
-        set_user_timeout(&socket, config.user_timeout_ms);
+        socket.set_user_timeout(config.tcp.user_timeout_ms);
         if let Err(err) = self.poll.registry().reregister(&mut socket, token, Interest::READABLE) {
-            warn!(?err, %peer_addr, "couldn't register connected tcp stream");
+            warn!(?err, %peer, "couldn't register connected stream");
             let _ = socket.shutdown(Shutdown::Both);
             return false;
         }
```
vladimir-ea, line 508: is it correct to apply tcp config to all sockets in the group - are they all tcp?

crates/flux-network/src/stream/network.rs
```diff
@@ -416,37 +402,32 @@ impl NetworkState {
...
-                set_user_timeout(&socket, config.user_timeout_ms);
+                socket.set_user_timeout(config.tcp.user_timeout_ms);
                 if let Err(err) =
                     self.poll.registry().register(&mut socket, token, Interest::READABLE)
                 {
-                    warn!(?err, %peer_addr, "couldn't register accepted tcp stream");
+                    warn!(?err, %peer, "couldn't register accepted stream");
                     let _ = socket.shutdown(Shutdown::Both);
                     continue;
                 }
```
vladimir-ea, line 580: does calling these functions on unix domain sockets return `Ok`?

crates/flux-network/src/stream/endpoint.rs
Bronek: This is **remote** which in case of Unix-domain socket is not a thing IMO - it is anonymous. For the **published** socket (of any kind - TCP or Linux-domain) you want to read `pub enum Endpoint` above, which has `Unix(PathBuf)` i.e. equality would compare path.

crates/flux-network/src/stream/network.rs
Bronek: The operation is no-op on Unix-domain sockets - see crates/flux-network/src/stream/transport.rs

crates/flux-network/src/stream/network.rs
Bronek: Unix-domain sockets have their own `UnixStream::shutdown` which is routed through `TransportStream`

```Rust
    pub(crate) fn shutdown(&self, how: Shutdown) -> io::Result<()> {
        match self {
            Self::Tcp(socket) => socket.shutdown(how),
            Self::Unix(socket) => socket.shutdown(how),
        }
    }
```

Bronek: .... however for operations which trivially do not apply, we do return `Ok`, e.g.

```Rust
    pub(crate) fn set_keepalive(&self) -> io::Result<()> {
        match self {
            Self::Tcp(socket) => set_keepalive(socket),
            Self::Unix(_) => Ok(()),
        }
    }
```

crates/flux-network/src/stream/endpoint.rs
vladimir-ea: ok .. i just wonder about all remote unix Peers being logically equal / indistinguishable and what confusion that might result in.

PR #139, opened 2026-08-26
flux-network: schedule services and add HttpService

Bronek:
This PR builds on #138. It adds network-scheduled services and replaces `HttpNetwork` with `HttpService`, an HTTP protocol layer over one `ConnectionGroup`. Five commits.

## Service scheduling

A service owns one `ConnectionGroup` and is scheduled by its network. `StreamNetwork::drive(max_timeout, services, unclaimed_handler)` runs one iteration: it folds network and service deadlines into the poll timeout, routes events by group ownership, then ticks services in slice order. Ticks receive the time after the poll wait.

The scheduling trait is private. Public callers pass opaque `ServiceRef`s made by `as_service()`, keeping the scheduling interface closed while allowing more service implementations later.

Each call validates the complete service slice before doing work. Omitting or duplicating a service whose group is claimed is therefore a configuration error, independent of which events arrive. Events for unclaimed groups go to the supplied handler. `poll_with` remains available for networks containing only unclaimed groups.

## HttpService

`HttpService::new(&mut net, group, config)` claims a raw-framed group. The service owns HTTP parsing and connection state; transport and queue policy remain in `ConnectionGroupConfig`.

HTTP events are now pull-based. `next_event(&mut net)` returns lifecycle, request and response events from ready connections. A request borrows its bytes from the connection buffer and carries a `Responder` for an inline answer. Dropping the responder defers the answer to `HttpService::respond(&mut net, token, ...)`.

Pulling lets a tile cap its work per iteration without losing readiness. `tick` continues to report work while an event remains available, including after a caller stops pulling early.

Answered input is tracked with cursors. It stops counting against connection limits immediately, while the buffer is compacted only when the consumed prefix reaches half its length. Inline answers reclaim on the next pull or network tick, after the event borrow ends; by-token answers hold no such borrow and reclaim before returning. A benchmark covers the compaction policy.

`close(self, &mut net)` closes the group and returns it empty and unclaimed. If a service is dropped while its network continues to run, the next driver call reports the missing owner. Dropping the service and network together during teardown remains harmless.

## Verification

Integration tests cover HTTP behavior through the public API over TCP and Unix sockets. Private unit tests pin cursor accounting, compaction, ready-queue behavior and the different reclamation points of inline and by-token answers without exposing representation accessors. Driver tests cover group claims, validation, deadline folding, event routing, tick order, post-poll time and did-work reporting.

One commit fixes the pre-existing `queued_messages_flush_on_second_connection_after_backpressure` race. The test previously waited five seconds for a marker that can arrive at that boundary under load. It now waits for an explicit marker from each collector, with a deadline only to bound failure.

Two review follow-ups document and test the answer-path reclamation rule, then move cursor and ready-queue assertions into private unit tests. `HttpService` exposes no representation accessors for testing.

## Scope

Driving from a caller-owned poll follows in A4. `listen` still returns `io::Result<()>`; reporting its bound endpoint follows later in the stack.

## Known test flakes

`stream_network_client_is_wire_compatible_with_tcp_connector_server` and the new loopback unit tests still guess unused TCP ports. That race is fixed later when `listen` reports the endpoint it bound. If one fails in CI, rerun the job.

Base: #138. Next: A4, driving the network from a caller-owned poll.

Refer: #143

Assisted-by: Codex:gpt-5

### DIFF
--- a/crates/flux-network/src/stream/network.rs
+++ b/crates/flux-network/src/stream/network.rs
@@ -10,7 +10,7 @@ use mio::{Events, Interest, Poll, Registry, Token, event::Event};
 use tracing::{debug, error, info, warn};
 
 use super::{
-    Endpoint, Peer, TcpTelemetry, set_socket_buf_size,
+    Endpoint, Peer, ServiceRef, TcpTelemetry, set_socket_buf_size,
     tcp_stream::{
         DEFAULT_TCP_USER_TIMEOUT_MS, FRAME_HEADER_SIZE, frame_payload_len, write_frame_header,
         write_frame_len, write_frame_ts,
@@ -266,7 +266,7 @@ impl ConnectionGroupConfig {
     }
 }
 
-/// Event emitted by [`StreamNetwork::poll_with`].
+/// Event emitted by [`StreamNetwork::drive`].
 pub enum StreamEvent<'a> {
     /// A listener accepted a new connection.
     Accepted { group: ConnectionGroup, token: Token, peer: Peer },
@@ -278,6 +278,18 @@ pub enum StreamEvent<'a> {
     Message { group: ConnectionGroup, token: Token, payload: &'a [u8], send_ts: Nanos },
     /// An established connection was closed.
     Disconnected { group: ConnectionGroup, token: Token, peer: Peer },
+}
+
+impl StreamEvent<'_> {
+    /// The group whose service, or whose unclaimed handler, this event is for.
+    pub(crate) fn group(&self) -> ConnectionGroup {
+        match *self {
+            Self::Accepted { group, .. } |
+            Self::Connected { group, .. } |
+            Self::Message { group, .. } |
+            Self::Disconnected { group, .. } => group,
+        }
+    }
 }
 
 struct GroupState {
@@ -437,8 +449,10 @@ impl NetworkState {
         self.connections[index].state = ConnectionState::Connecting(socket);
     }
 
-    fn maybe_reconnect(&mut self) {
-        let now = Instant::now();
+    /// Retries every outbound endpoint whose group is due one, reporting
+    /// whether an attempt was made.
+    fn maybe_reconnect(&mut self, now: Instant) -> bool {
+        let mut attempted = false;
         for group_index in 0..self.groups.len() {
             if !self.groups[group_index].reconnector.fired_at(now) {
                 continue;
@@ -449,9 +463,26 @@ impl NetworkState {
                     matches!(self.connections[index].state, ConnectionState::Disconnected)
                 {
                     self.start_connect(index);
+                    attempted = true;
                 }
             }
         }
+        attempted
+    }
+
+    /// The earliest instant the network's own timers need a call at: the next
+    /// retry of a group holding an outbound endpoint that is down.
+    fn next_deadline(&self) -> Option<Instant> {
+        let mut next: Option<Instant> = None;
+        for connection in &self.connections {
+            if connection.endpoint.is_some() &&
+                matches!(connection.state, ConnectionState::Disconnected)
+            {
+                let fire = self.groups[connection.group.0].reconnector.next_fire();
+                next = Some(next.map_or(fire, |next: Instant| next.min(fire)));
+            }
+        }
+        next
     }
 
     fn finish_connect<F>(&mut self, index: usize, handler: &mut F) -> bool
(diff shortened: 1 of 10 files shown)


crates/flux-network/src/stream/network.rs
```diff
@@ -283,8 +295,30 @@ impl NetworkState {
...
+    ///
+    /// Closing an [`Endpoint::Unix`] listener unlinks its socket file.
+    fn close_group(&mut self, group: ConnectionGroup) {
+        for index in (0..self.listeners.len()).rev() {
+            if self.listeners[index].group == group {
+                let mut listener = self.listeners.swap_remove(index);
+                let _ = self.poll.registry().deregister(&mut listener.socket);
+            }
+        }
+        for index in (0..self.connections.len()).rev() {
```
vladimir-ea, line 463: why `rev()`?

vladimir-ea: so you swap remove at the end? does it make a difference?

crates/flux-network/src/stream/service.rs
```diff
@@ -0,0 +1,422 @@
...
+        /// The earliest instant this service needs a tick at.
+        fn next_deadline(&self) -> Option<Instant>;
+    }
+}
+
+/// One service, for the duration of a driver call.
+///
+/// A service hands its network a `ServiceRef` — `HttpService::as_service()` —
+/// and the network drives it; the carrier itself offers a caller nothing.
+pub struct ServiceRef<'a>(&'a mut dyn private::ServiceDriver);
```
vladimir-ea, line 36: we should avoid dynamic dispatch

crates/flux-network/src/stream/network.rs
```diff
@@ -784,22 +877,135 @@ impl StreamNetwork {
...
+    /// in the same iteration rather than the next.
+    ///
+    /// # Panics
+    /// Every service-owned group must appear exactly once among `services`. A
+    /// group whose service is missing or doubled is a configuration error and
+    /// panics before the call does anything else.
+    pub fn drive<F>(
+        &mut self,
+        max_timeout: Option<Duration>,
+        services: &mut [ServiceRef<'_>],
```
vladimir-ea, line 1133: why are services pushed down into the network here? this seems somewhat entangled. the previous api passed in a closure to handle events, why not stick with that approach? multiplexing to different services can live in the layer above?

vladimir-ea: also, this change from `poll_with` to `drive` breaks the existing api - which is I think the ocncern @ltitanb has raised - is it strictly necessary?

vladimir-ea: actually I see that `poll_with` is preserved below .. 👍

crates/flux-network/src/stream/network.rs
```diff
@@ -784,22 +877,135 @@ impl StreamNetwork {
...
+    pub fn poll_with<F>(&mut self, handler: F)
+    where
+        F: for<'a> FnMut(StreamEvent<'a>),
+    {
+        if let Some(index) = self.claimed.iter().position(|claimed| *claimed) {
+            panic!(
+                "connection group {index} has a service — drive the network with \
+                 StreamNetwork::drive, passing every service"
+            );
         }
```
vladimir-ea, line 1193: maybe make this whole check a `debug_assert!` ?

crates/flux-network/src/stream/network.rs
```diff
@@ -784,22 +877,135 @@ impl StreamNetwork {
...
     where
         F: for<'a> FnMut(StreamEvent<'a>),
     {
-        self.state.drain_pending_disconnects(&mut handler);
-        self.state.maybe_reconnect();
-        if let Err(err) = self.state.poll.poll(&mut self.events, Some(std::time::Duration::ZERO)) {
-            if err.kind() != io::ErrorKind::Interrupted {
-                flux_utils::safe_panic!("couldn't poll the stream network: {err}");
+        let now = Instant::now();
+        self.validate(services);
```
vladimir-ea, line 1140: this validate on every tick seems excessive cost

crates/flux-network/src/stream/network.rs
vladimir-ea: but it still seems like a cleaner change to handle services and service dispatch above the network rather than within it? i.e. inside the handler passed to `poll_with`?

crates/flux-network/src/stream/network.rs
```diff
@@ -846,6 +1052,31 @@ impl StreamNetwork {
...
+        service.on_event(&event);
+    } else {
+        debug_assert!(
+            !claimed[group.0],
+            "service-owned group {} has no service to route to",
+            group.0
+        );
+        unclaimed_handler(event);
+    }
+}
```
vladimir-ea, line 1343: where does the requirement to have multiple `Service`s registering an interest with a single network instance come from? is there a use case?

crates/flux-network/src/stream/network.rs
Bronek: it does, for correctness, and it keeps the each removal bound at O(1)

crates/flux-network/src/stream/service.rs
Bronek: Indeed - the refactoring is already being prepared.

crates/flux-network/src/stream/network.rs
Bronek: The cost is proportional to the number of services, and there shouldn't be more than a few - so it's very cheap.

crates/flux-network/src/stream/network.rs
Bronek: One network per service would mean one poll per service. We want to allow one-poll-per-tile model (which in turn can feed several services) - this is why we support the External poll.

crates/flux-network/src/stream/network.rs
vladimir-ea: I understand the swap_remove is O(1) was wondering why reverse the iteration?

PR #141, opened 2026-08-27
flux-network: drive the stream network from a caller-owned poll

Bronek:
This PR builds on #139. It adds the External poll mode from ADR 0001: a `StreamNetwork` that registers on the caller's poll and never polls itself. One commit.

## Poll ownership

Poll ownership is fixed when the network is constructed. `StreamNetwork::default()` creates an Owned network driven by one `drive` call per iteration. `StreamNetwork::with_registry(registry, token_base)` creates an External network driven through the caller's poll:

```rust
let mut net = StreamNetwork::with_registry(poll.registry().try_clone()?, Token(1024));

let timeout = net
    .next_deadline(&services)
    .map(|deadline| deadline.saturating_sub(Instant::now()).into());
poll.poll(&mut events, timeout)?;

for event in &events {
    let ours = net.handle_event(event, &mut services, |_| {});
    worked |= ours;
    if !ours {
        // Handle a source registered by the caller.
    }
}
worked |= net.tick(&mut services, |_| {});
```

Each mode has one driving interface. Owned networks refuse `next_deadline`, `handle_event` and `tick`; External networks refuse `drive`, `poll_with` and `waker`.

Both modes deliver readiness and maintenance events before ticking services. Owned runs maintenance before its poll; External handles the poll's events first and runs maintenance in `tick`. Ordering across different connections may therefore differ and is not a correctness guarantee.

## Tokens and wakers

An External network allocates tokens upward from `token_base` without reuse. `handle_event` returns `true` for a token in that range, including a stale token whose connection has closed, and `false` for a caller-owned source. Folding this ownership result into a work signal is conservative: stale or otherwise non-producing readiness may count as one busy iteration.

Caller-owned tokens stay below the base. One External network is supported per poll because tokens above its current high-water mark cannot be distinguished from future allocations.

An Owned network can provide one `mio::Waker` on the reserved token `Token(usize::MAX)`. Waking it interrupts a blocked `drive` without counting as work. External callers create a waker on their own poll and token.

`flux_network::mio` re-exports the mio types used by the network, so callers can construct the exact `Poll`, `Registry`, `Event` and `Waker` types required by the API.

## Verification

External-mode tests cover the three-call loop, deadline folding, reconnects, foreign and stale events, validation order, mode misuse, wakers, token bounds and disconnect delivery. The HTTP suite runs its baseline behavior over Owned and External polls, with both TCP and Unix sockets.

The loopback tests, this PR's included, still guess unused TCP ports. That race is fixed later in the stack when `listen` reports the endpoint it bound; rerun the job if it appears in CI.

Workspace tests, Clippy with warnings denied, the nightly formatting check and `cargo doc` pass at this tip with no new warnings.

Base: #139. Next: A5, connection admission and half-close.

Refer: #143

Assisted-by: Codex:gpt-5

### DIFF
--- a/crates/flux-network/src/lib.rs
+++ b/crates/flux-network/src/lib.rs
@@ -1,4 +1,8 @@
 pub mod http;
 
 pub mod stream;
+/// The mio a network is built on. An External-mode caller reaches `Poll`,
+/// `Registry`, `Waker`, `Events` and `event::Event` through here, so the poll
+/// it drives and the sockets flux registers on it are the same mio.
+pub use mio;
 pub use mio::Token;
--- a/crates/flux-network/src/stream/mod.rs
+++ b/crates/flux-network/src/stream/mod.rs
@@ -1,3 +1,93 @@
+//! Poll-driven byte-stream networking: TCP and Unix-domain sockets in one
+//! [`StreamNetwork`], with protocol layers as services inside it.
+//!
+//! # Poll ownership
+//! A tile hosting sockets owns exactly one poll, so which side holds it is a
+//! construction-time mode of the network.
+//!
+//! **Owned** — [`StreamNetwork::default`] — creates a poll of its own and
+//! drives it: one call per iteration, [`StreamNetwork::drive`], runs
+//! maintenance, folds every deadline, polls once, routes each event to the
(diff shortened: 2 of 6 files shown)


louisponet:
Could we also get some pre and after performance benchmarks and impact?

PR #149, opened 2026-09-03
feat(network): add batched TCP sends

ltitanb:
## Summary

- add `send_many_with` and `broadcast_many_with` for batched TCP writes
- isolate each serialized frame with `PayloadBuf`
- preserve length-prefixed frame boundaries and concatenate raw payload batches
- support standard writers and Wincode serialization
- bump the workspace version to `0.2.0` for the public API change

## Load test

Both processes used spare CPU 4. The public-network RTT was approximately 0.8 ms.

Each message was 1,280 bytes. Each case ran three times for five seconds.
The table contains median sender CPU time.

| Batch | Rate | Flux 0.1.3 single | Flux 0.2.0 single | Flux 0.2.0 batch | Batch vs 0.1.3 |
| ---: | ---: | ---: | ---: | ---: | ---: |
| 1 | 4k msg/s | 575 ms | 572 ms | 572 ms | -0.6% |
| 4 | 16k msg/s | 762 ms | 636 ms | 583 ms | -23.5% |
| 8 | 32k msg/s | 980 ms | 939 ms | 582 ms | -40.6% |

All runs delivered every message in order. No send was rejected. No payload was corrupted.

## Validation

- `just fmt-check`
- `cargo check --workspace --all-features`
- `just clippy`
- `cargo test --workspace --all-features --locked`
- `python3 .github/scripts/test_release.py`
- `python3 .github/scripts/release.py`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -585,7 +585,7 @@ checksum = "5baebc0774151f905a1a2cc41989300b1e6fbb29aff0ceffa1064fdd3088d582"
 
 [[package]]
 name = "flux"
-version = "0.1.3"
+version = "0.2.0"
 dependencies = [
  "bitcode",
  "core_affinity",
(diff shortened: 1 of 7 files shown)


niki4smirn (approved):
lgtm

PR #154, opened 2026-09-05
add truncate and rename file ops

vladimir-ea:
also add optional user provide operation ids for sync type operations so user can track when they complete

**Breaking changes**
- adds 2 new `DiskEvent` variants: `Truncated` and `Renamed` which will BREAK exhaustive enum matches
- adds `Option<OperationId>` field to `Synced` and `Failed` events

### DIFF
--- a/crates/flux-disk/src/lib.rs
+++ b/crates/flux-disk/src/lib.rs
@@ -46,18 +46,15 @@ const EMPTY_PATH: &[u8] = b"\0";
 #[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
 pub struct FileToken(usize);
 
-/// Caller-chosen identifier echoed by tagged completion and failure events.
+/// Identifies an accepted sync, truncate or rename operation.
 ///
-/// `DiskIo` does not interpret this value or require it to be unique. It is
-/// intended for correlating durability barriers with higher-level work.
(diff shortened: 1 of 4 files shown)


crates/flux-disk/src/lib.rs
```diff
@@ -397,20 +474,71 @@ impl DiskIo {
...
+    /// Later operations do not start until the truncate completes. The append
+    /// cursor is unchanged; use [`Self::set_write_cursor`] when repairing an
+    /// append-only file. Requires Linux 6.9 or newer.
+    pub fn truncate(&mut self, file: FileToken, len: u64) -> bool {
+        self.push_truncate(file, len, None)
+    }
+
+    /// Like [`Self::truncate`], echoing `operation_id` in its completion or
+    /// failure event.
+    pub fn truncate_with_id(
```
louisponet: I wonder if we could just have 1 function e.g. `trunctate -> Option<OperationId>` which creates the operationid by itself? That way we can remove all the optionality and I think it also cleans up all the `_with_id` functions.

vladimir-ea: and change return types of `sync_all` and `sync_data` also? so `None` where we previously has `false`?

louisponet: Yeah that's what I was thinking

louisponet (approved):

PR #155, opened 2026-09-07
feat(network): UDP publish/subscribe feed

Sham-BahamasTrading:
Adds flux_network::udp, a one-to-many UDP feed alongside the existing TCP transport

Flow
 1. Dial. A subscriber sends an 8-byte SUBSCRIBE to the publisher
 2. Publish. publish() splits the message into 1200-byte datagrams and sends each to every target. Returns the sequence number, or None when 0 subscribers.
 3. Receive. poll_with(handler) drains the socket and calls the handler once per whole message, reassembling fragments per publisher.

Wire format:
  32-byte header: magic, version, session id, sequence, length, offset, send time, payload.
  Total 1200 bytes

### DIFF
--- a/crates/flux-network/src/udp/subscriber.rs
+++ b/crates/flux-network/src/udp/subscriber.rs
@@ -1,0 +1,173 @@
+use std::{
+    io,
+    net::{SocketAddr, UdpSocket},
+};
+
+use flux_communication::Timer;
+use flux_timing::Nanos;
+use rustc_hash::FxHashMap;
+
+use super::wire::{
+    DEFAULT_MAX_DATAGRAM_SIZE, Fragment, MAX_DATAGRAM_SIZE, SUBSCRIBE, UDP_HEADER_SIZE,
+    fragment_count,
+};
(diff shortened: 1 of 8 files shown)


louisponet (commented):
Let's do the same thing as we do for tcp connector no? Use epoll and have the same api/structure?

crates/flux-network/src/udp/subscriber.rs
```diff
@@ -0,0 +1,173 @@
...
+            }
+
+            let complete;
+            let total = fragment_count(header.len as usize, FRAGMENT_PAYLOAD_SIZE);
+            let payload = if total == 1 {
+                fragment.payload
+            } else {
+                let Some(buf) = source.assemble(&fragment, total) else { continue };
+                complete = buf;
+                &complete[..]
```
vladimir-ea, line 129: since there is no ack / resend, any lost packet from a multi-fragment message will result in a `partial` sitting in the source map indefinitely.  maybe they could have an expiry?

PR #156, opened 2026-09-08
feat(network): reliable UDP transport in NetworkDriver

gd-0:
## Note to reviewer:
There were no logic changes to TCP. we have a new wrapper NetworkDriver struct with the identical API that takes both transports. You will see changes to the TCP connector because:
- had to move code out because of the new wrapper
- we now have a shared config struct for shared config between UDP and TCP which means that some of the *inner* (never called externally) poll etc fns inside TCP now take the config param 

Can essentially skip all changes to tcp. Start at `NetworkDriver` to see how the wrapper works, then dig into the UDP impl

## What

`TcpConnector` becomes `NetworkDriver` with `Transport::{Tcp(TcpConfig), Udp(UdpConfig)}`. Same API, same events, same poll loop; callers import `flux_network::NetworkDriver` and pick the transport.

`network_driver.rs` dispatches to `tcp/connector.rs` (the old code, config passed by reference) or `udp/`.

### How the UDP transport works

- A message is split into datagrams. Each carries a session id, a sequence number, the message length and its fragment index. Receiver reassembles and delivers each message the moment it is complete, in whatever order messages complete.
- Receiver acks after each batch it reads: everything received up to a point, plus a bitmap of what it has beyond that. Sender resends anything the receiver has visibly skipped. If acks stop entirely, it resends the oldest unacked datagram and grows from there. Retransmit timeout adapts to the measured round trip; `UdpConfig::lan()` / `wan()` set the bounds.
- Broadcast serialises the message once. All fragments for all peers on a socket go out in one `sendmmsg`. Whatever the kernel does not take is picked up on the next writable event. Each peer keeps its own unacked window pointing into the shared copy, released when every peer has acked.
- Connections handshake with a hello and get a fresh session per attempt. A restarted or dropped side is told to reconnect immediately rather than waiting for the timeout. On reconnect, unacked messages are replayed whole, so delivery across a reconnect is at-least-once.
- `with_max_backlog` counts unacked datagrams, `with_user_timeout` is the silence timeout. A message that cannot fit the send window disconnects the peer.
- There is no flow control yet. The receiver's socket buffer has to hold a full send window (`with_socket_buf_size` plus `net.core.rmem_max`). First follow-up.

## Measurements

Loopback, sender and receiver on separate pinned cores, one-way p50 (`benches/udp_pipeline.rs`):

| | TCP | UDP |
|---|---|---|
| 2 KB, one receiver | 3.6 µs | 4.7 µs |
| 2 KB broadcast to 8 receivers | 243 µs | 18.5 µs |
| 2 MiB, one receiver | 172 µs | 2865 µs |

- Single small message to one peer: about the same.
- Small messages to many subscribers (the RPC to builder shape): 13x faster, one syscall for all peers instead of one write per peer.
- Large messages: worse. The kernel pays about 1.6 µs per datagram and batching does not remove it. UDP GSO would; one for another day.
- One lost datagram in a 2 MiB message costs one retransmit, not the message.

## Integration

```rust
let mut connector = NetworkDriver::default()
    .with_transport(Transport::Udp(UdpConfig::lan()))   // wan() for cross-region feeds
    .with_socket_buf_size(32 << 20)
    .with_max_backlog(MAX_SUBSCRIBER_BACKLOG, Duration::from_secs(MAX_SUBSCRIBER_BACKLOG_FOR_SECS));
connector.listen_at(addr);
connector.write_or_enqueue_with(SendBehavior::Broadcast, |b| b.extend_from_slice(payload));

let connector = NetworkDriver::default()
    .with_transport(Transport::Udp(UdpConfig::lan()))
    .with_socket_buf_size(32 << 20)
    .with_user_timeout(USER_TIMEOUT_MS);
connector.connect(addr);            // token semantics as today
connector.poll_with(|event| ...);   // Message / Reconnect / Disconnect as today
```

Msgs arrive out of order; nothing in the subscriber depends on order. Raise `net.core.rmem_max` to the buffer size on both boxes.

## Tests

13 UDP integration tests: roundtrip, broadcast, exactly-once delivery through a 1-in-7 lossy relay, reconnect replay of partly acked messages, client and server restart, server-side reset, window exhaustion, dropped peer mid-broadcast, dcache. Unit tests for the window, ack bitmap, stale acks and rewind.

## Follow-ups

Receive or congestion window. `UDP_SEGMENT` for large messages. `TcpNetwork` duplicates most of `tcp/connector.rs` and should fold into it.


gd-0:
@clanky-gattaca review

crates/flux-network/src/tcp/connector.rs
```diff
@@ -333,25 +257,25 @@ impl ConnectionManager {
         }
     }
 
-    fn connect(&mut self, addr: SocketAddr) -> Option<Token> {
+    pub(crate) fn connect(&mut self, cfg: &Config, addr: SocketAddr) -> Option<Token> {
```
louisponet: not loving that config is now a parameter everywhere. Feels like this should be owned by the connection that was created with a given config

crates/flux-network/src/tcp/connector.rs
```diff
@@ -587,349 +519,92 @@ impl ConnectionManager {
...
-    /// Writable events trigger retries of queued writes. Because this method
-    /// performs a single non-blocking poll, calling it infrequently can slow
-    /// backlog draining under backpressure, particularly when using small
-    /// socket buffers configured via [`Self::with_socket_buf_size`].
+    /// Delivers pending disconnects and reconnects, then polls once.
     #[inline]
-    pub fn poll_with<F>(&mut self, mut handler: F) -> bool
+    pub(crate) fn poll_with<F>(
+        &mut self,
+        cfg: &Config,
```
louisponet: let's not change the api here at all, don't think it's needed. We should change the lower levels to accept Poll that could be shared between different impls if needed rather than change this toplevel api wrapper imo

crates/flux-network/src/udp/connector.rs
```diff
@@ -0,0 +1,661 @@
...
+        let now = Instant::now();
+        for peer in self.peers.iter_mut().filter(|p| p.is_outbound() && !p.is_connected()) {
+            let entry = &self.sockets[socket_index(&self.sockets, peer.token)];
+            peer.send_hello(&entry.socket, now);
+        }
+    }
+
+    /// Hello retries, retransmits, heartbeats and peer timeouts.
+    fn tick(&mut self, cfg: &Config, now: Instant) {
+        let peer_timeout = Duration::from_millis(u64::from(cfg.user_timeout_ms));
```
louisponet: this seems sub optimal?

crates/flux-network/src/udp/peer.rs
```diff
@@ -0,0 +1,1320 @@
...
+    fn is_acked(&self, seq: u64) -> bool {
+        let i = seq & self.mask;
+        self.acked[(i / 64) as usize] & (1 << (i % 64)) != 0
+    }
+
+    #[inline]
+    fn set_acked(&mut self, seq: u64, v: bool) {
+        let i = seq & self.mask;
+        let bit = 1 << (i % 64);
+        let word = &mut self.acked[(i / 64) as usize];
```
louisponet: isn't / 64 some kind of bitshift, same with %?

crates/flux-network/src/udp/peer.rs
```diff
@@ -0,0 +1,1320 @@
...
+    pub(crate) fn tick(
+        &mut self,
+        store: &MsgStore,
+        socket: &UdpSocket,
+        batch: &mut SendBatch,
+        now: Instant,
+        peer_timeout: Duration,
+    ) -> Result<SendOutcome, ()> {
+        if !self.connected {
+            let interval = Duration(
```
louisponet: wtf am I reading here, can't that be set once per connection some of this?

crates/flux-network/src/udp/peer.rs
```diff
@@ -0,0 +1,1320 @@
...
+    ctrl: Vec<u8>,
+}
+
+impl UdpPeer {
+    pub(crate) fn new(
+        addr: SocketAddr,
+        token: Token,
+        socket_token: Token,
+        local_session: u32,
+        config: &UdpConfig,
```
louisponet: still think this should just be part of the peer no? feels like a lot of cleanup could be achieved

louisponet (approved):

crates/flux-network/src/tcp/connector.rs
gd-0: Agreed. Each manager now owns its Config (2abb90e). No cfg parameters left; NetworkDriver is dispatch plus the scratch buffer, and the with_* builders write into whichever manager is active.

crates/flux-network/src/tcp/connector.rs
gd-0: Reverted to poll_with(handler). Each manager owns its Poll and Events again, as TcpConnector did; nothing shares a poll today so the top-level API stays as it was.

crates/flux-network/src/udp/connector.rs
gd-0: Converted once now: Config stores the timeout as a Duration, set in the builder. TCP converts back to ms only when it sets the socket option.

crates/flux-network/src/udp/peer.rs
gd-0: Same codegen on unsigned, but written as >> 6 and & 63 now so nobody has to check.

crates/flux-network/src/udp/peer.rs
gd-0: The peer stores hello_due when it sends a hello (backoff applied there); the tick is one Instant compare.

crates/flux-network/src/udp/peer.rs
gd-0: Done, UdpPeer owns its UdpConfig instead of copying four fields out of it.

PR #159, opened 2026-09-14
Versioned blobs, persistence naming, and SQL schemas in flux-versioned-types

louisponet:
Brings the builder telemetry machinery that is generic over metadata into flux: VersionedBlob (single latest on-disk format), InternalMetadata, TelemetrySchema plus its derive, and versioned_telemetry macros that always emit schemas with a name-derived, overridable persist directory.


crates/flux-versioned-types/src/lib.rs
```diff
@@ -41,14 +50,52 @@ macro_rules! versioned_struct {
...
+/// [`TelemetrySchema`] so it is queryable, and the type gets a
+/// [`VersionedPersistable`] home: the type name by default, or
+/// `persist = "dir"` when given.
+#[macro_export]
+macro_rules! versioned_telemetry {
+    ($name:ident, persist = $dir:expr => $($tokens:tt)*) => {
+        $crate::__versioned_telemetry_inner!($name, $dir => $($tokens)*);
+    };
+    ($name:ident => $($tokens:tt)*) => {
+        $crate::__versioned_telemetry_inner!($name, stringify!($name) => $($tokens)*);
```
alexX512: Might be better don't allow to have persistable without manual specification of dir name? Because in our repos we sometimes have same name in different namesapaces like `bundle::X`, `transaction::X` and `stringify!($name)` will produce only `X`

crates/flux-versioned-types/src/lib.rs
```diff
@@ -41,14 +50,52 @@ macro_rules! versioned_struct {
...
+    };
+}
+
+#[macro_export]
+#[doc(hidden)]
+macro_rules! __versioned_telemetry_inner {
+    ($name:ident, $dir:expr => $($tokens:tt)*) => {
+        $crate::evolve_struct! {
```
alexX512, line 112: Don't we use enums somewhere for versioned_telemetry? Because here is only struct impl

crates/flux-versioned-types/src/blob.rs
```diff
@@ -0,0 +1,371 @@
...
+
+impl VersionedBlob {
+    /// Pack serializable data and metadata vectors, tagging them with `T`'s
+    /// version so [`VersionedBlob::data_as`] can migrate them later.
+    pub fn from_parts<T: TypeHash, Data: Serialize, Metadata: Serialize>(
+        data: &Data,
+        metadata: &Metadata,
+    ) -> Self {
+        Self {
+            type_hash: T::TYPE_HASH ^ 123_456,
```
alexX512, line 106: Might be we can move that 123456 const to variable to use it in both encode/decode places? Tbh don't know why we even use it if hash already random

alexX512 (approved):

PR #161, opened 2026-09-14
TypeHash for ArrayStr

louisponet:
Sibling of the existing ArrayVec impl: hash name, capacity, and layout. Needed so versioned structs can hold ArrayStr fields without literals.


louisponet:
Folded into #160 (commit 4087e32).

PR #168, opened 2026-09-17
feat(tile): pin tiles to a set of cores

louisponet:
TileConfig stores cores: Vec<usize> instead of a single Option<usize>, with new_on_cores/background_on_cores constructors and a cores() accessor. Existing new/background constructors delegate unchanged.

tile_runner pins via sched_setaffinity over the whole set (empty means unpinned); thread_boot now takes &[usize]. Non-Linux falls back to pinning the first core.

Verified: just fmt, just clippy (zero warnings), cargo test -p flux --all-features, cargo check --workspace --all-features.


crates/flux-utils/src/thread.rs
```diff
@@ -47,9 +48,32 @@ fn set_thread_niceness(niceness: Option<ThreadNiceness>) {
...
 
-fn set_thread_affinity(core: usize) {
-    if !core_affinity::set_for_current(CoreId { id: core }) {
-        warn!(?core, "couldn't set core affinity");
+#[cfg(target_os = "linux")]
+fn set_thread_affinity(cores: &[usize]) {
+    if cores.is_empty() {
+        return;
+    }
+    // ponytail: raw sched_setaffinity because core_affinity only pins a single core
```
gd-0: comment

crates/flux/src/tile/mod.rs
```diff
@@ -14,23 +14,38 @@ use crate::{
...
 
 impl TileConfig {
     pub fn new(core: usize, thread_niceness: Option<ThreadNiceness>) -> Self {
-        Self { core: Some(core), thread_niceness, min_loop_duration: None, metrics: true }
+        Self::new_on_cores(vec![core], thread_niceness)
+    }
+
+    /// Pin the tile to a set of cores instead of a single core.
+    pub fn new_on_cores(cores: Vec<usize>, thread_niceness: Option<ThreadNiceness>) -> Self {
+        Self { cores, thread_niceness, min_loop_duration: None, metrics: true }
```
gd-0: this fn should only be there for the `background` variant

gd-0 (approved):

PR #172, opened 2026-09-17
feat(versioned-types): per-version wire_skip; ByteStable for floats

louisponet:
Applications migrating telemetry types onto the byte-stable wire have old versions that were only ever bincode-encoded. Many are padded (a `bool` next to a `u64`) or carry data in enum variants, so they can never be cast from bytes. Until now the macros emitted the `ByteStable` derive on every version of a chain, and the only opt-out was the whole chain. Those old versions cannot be edited either: adding `_pad` changes their locked type hash and breaks decoding of the files they wrote.

## What changes

- A version block may carry `#[wire_skip]` next to its `#[type_hash_lock]`. A skipped version gets no `ByteStable` derive, `version_size` returns `None` for its hash, `decode_versions` reports `UnknownTypeHash`, and `VERSION_HASHES` lists only the versions that can appear in a blob. The bincode path (`versioned_deserialize_vec`) still migrates it exactly as before, so old files keep decoding.
- The newest version of a chain cannot be skipped (a chain-level `#[wire_skip]` is the way to keep a type off the wire entirely), and a version-level skip on an already skipped chain is rejected. Both are compile errors on the attribute.
- `f32` and `f64` are `ByteStable`: every bit pattern is a valid float, NaNs included.

```rust
versioned_struct!(Foo =>
    #[wire_skip]                              // padded, bincode only
    #[type_hash_lock(hash = ..)]
    FooV1 { pub a: u8, pub b: u64 }
    #[type_hash_lock(hash = ..)]
    FooV2 { modify { a: u64 = |a: u8| u64::from(a) } }
);
```

`crates/flux-versioned-types/tests/version_wire_skip.rs` checks, on a struct chain and an enum chain, that the skipped version is off the wire, that the bincode path still migrates it, and that the newest version round-trips through a `BlobCache`.

Review guide — where the lines are (additions/deletions per file):

| area | file | lines |
|---|---|---|
| macro | [shared/generate.rs](https://github.com/gattaca-com/flux/pull/172/files#diff-2a2a5f15c1dd60ccfe1da4ddb0b94036b73bfbb9a80dd9cac8b07af588d40bd7) | +61/−3 |
| macro | [rolling/generate.rs](https://github.com/gattaca-com/flux/pull/172/files#diff-c610c6cb82954e8a4f5a199bf687cc0e39717ccada581d06c41ca0579be35355) | +19/−4 |
| floats | [byte-stable/src/lib.rs](https://github.com/gattaca-com/flux/pull/172/files#diff-61e9b737c641d374a27cac02bb9630d26f327672526a44d678d77b9a6c6145b9) | +1/−1 |
| behaviour | [version_wire_skip.rs](https://github.com/gattaca-com/flux/pull/172/files#diff-6ef6ac63bbb1c8fc110b893fbd07571e298adc1c19e0e82dd3a8d3e8f3b85a2e) | +68/−0 |

<details>
<summary>Supporting: evolve, evolve_enum, rolling, flux-versioned-types (+21/−7)</summary>

| area | file | lines |
|---|---|---|
| evolve | [evolve/generate.rs](https://github.com/gattaca-com/flux/pull/172/files#diff-90d3c5c71465a996e15a92b3b527870025ab333c536d827b2fe0896fcf946c7c) | +5/−2 |
| evolve_enum | [evolve_enum/generate.rs](https://github.com/gattaca-com/flux/pull/172/files#diff-100f550d95751280cedd43711619ddcf8f821b336ac5b754794b4ba46ed16f81) | +5/−2 |
| rolling | [mod.rs](https://github.com/gattaca-com/flux/pull/172/files#diff-82723186dd96756cbdde950603e7c5ae535f622e71d2891021c012ad42d7d8d3) | +7/−1 |
| flux-versioned-types | [flux-versioned-types/src/lib.rs](https://github.com/gattaca-com/flux/pull/172/files#diff-b0829ccda356745ebd4c9784fa976b6b1d86c880ded8400b8cac44b3316f0696) | +4/−2 |
</details>


gd-0 (approved):
lot of tests that can prob go btw

PR #176, opened 2026-09-17
feat(network): Postgres client and HTTP layer on one shared poll

louisponet:
A tile that writes to both Postgres and ClickHouse should wait on all of its sockets in one poll. Until now `HttpNetwork`, which the ClickHouse client rides on, owned its own `TcpNetwork` and with it its own epoll instance, so a second database meant a second poll per loop iteration and no shared wait. This PR adds `flux-postgres`, a Postgres wire-protocol client that runs inside a tile's existing loop with no async runtime or thread, and moves `HttpNetwork` onto the same caller-owned network, so Postgres, ClickHouse and the tile's raw TCP groups share one poll.

## One network, several protocol layers

Every protocol layer has the same shape over a `TcpNetworkCore`:

- `on_event(&TcpEvent) -> bool` runs inside the network's `poll_with` handler and takes the events of the layer's own TCP group; events of other groups are left for the next layer.
- `drive(&mut TcpNetworkCore, handler)` delivers the layer's own events, sweeps timeouts, and sends queued requests on idle connections.
- Connections open explicitly: `Postgres::connect(&mut net)` after the builders, `HttpNetwork::listen`/`connect`/`pool`, and `ClickHouse::new(http, net, addr, n)` for its pool. `close(self, net)` removes a layer's endpoints, since `disconnect` alone lets them reconnect.
- `TcpNetwork` and `TcpNetworkWithExternalPoll` both coerce to `TcpNetworkCore` at the call site, so a layer runs under either poll owner without generics.

```rust
net.poll_with(|event| {
    if !http.on_event(&event) {
        pg.on_event(&event);
    }
});
http.drive(&mut net, |event| {
    if let Some((id, result)) = ch.outcome(&event) { /* ClickHouse outcome */ }
});
pg.drive(&mut net, |id, result| { /* Postgres outcome */ });
```

`HttpNetwork::send` now only queues and the next `drive` sends, so `ClickHouse::query`/`insert` keep their signatures; only `ClickHouse::new` takes the network. Callers of `HttpNetwork` and `ClickHouse::new` must adapt, so the next release is a minor bump under the pre-1.0 policy.

## The Postgres client

- `query`, `copy`, `copy_rows`, and `copy_text_rows` queue requests inside the client and return a `QueryId` at once; `drive` sends them on idle pooled connections and delivers one outcome per id.
- A full queue refuses the new request instead of failing queued ones: `query` returns `None`, the copy calls hand their encoded body back. A request lost mid-flight reports `Disconnected` and the endpoint reconnects on its own; queued requests wait out a pool outage.
- Authentication covers trust, cleartext, and SCRAM-SHA-256. The startup message goes out through the network's `on_connect_msg`; COPY pipelines its statement, data, and trailer in one send.
- Replies are owned and capped by `max_output_bytes`; server errors carry the SQLSTATE code and message.

## Encoding a batch

Both encoders turn any `serde::Serialize` row into a COPY body and name the columns after the row's fields, quoted. Every row in a batch must serialize the same columns; a batch that does not is rejected before anything is sent, so a `skip_serializing_if` field can never land in a neighbouring column.

`copybinary` is the fast path, but a type only reaches it if it has a binary encoding. `copytext` writes the same rows in Postgres' text format, where a value's text form is the literal the server parses for its column: `chrono` timestamps, `BigDecimal` numerics, and unit enum variants work there and nowhere else, which is what the builder's telemetry rows are made of. Text costs the server a parse per value, so prefer binary when the row allows it.

```rust
use flux_network::tcp::TcpNetwork;
use flux_postgres::{Error, Postgres};

#[derive(serde::Serialize)]
struct Fill { slot: u64, hash: String, amount: u128 }

let mut net = TcpNetwork::default();
let mut pg = Postgres::new("10.0.0.5:5432".parse()?)
    .with_credentials("writer", &password)
    .with_database("telemetry")
    .with_connections(2);
pg.connect(&mut net);

loop {
    net.poll_with(|event| {
        if pg.on_event(&event) {
            return;
        }
        // events for the tile's other endpoints and listeners
    });
    pg.drive(&mut net, |id, result| match result {
        Ok(output) => tracing::debug!(?id, tag = output.tag, "copy committed"),
        Err(Error::Server { code, message }) => {
            tracing::error!(?id, %code, "copy rejected: {message}");
        }
        Err(other) => tracing::error!(?id, ?other, "copy failed"),
    });

    if let Some(fills) = slot_finished() {
        if pg.copy_rows("fills", &fills).is_err() {
            tracing::warn!("postgres queue full; batch dropped");
        }
    }
}
```

Not covered: TLS, the extended protocol, `COPY TO STDOUT`, `LISTEN`, cancellation, per-request timeouts.
Review guide — where the lines are (additions/deletions per file):

| area | file | lines |
|---|---|---|
| shared layer | [src/http.rs](https://github.com/gattaca-com/flux/pull/176/files#diff-f03f4dc4c7cb76e3208557ecf4eb15338052ad59775f3413703cb45811964ae8) | +124/−100 |
| postgres client | [flux-postgres/src/lib.rs](https://github.com/gattaca-com/flux/pull/176/files#diff-d3eab964732859136066cf078ba9c6c3e212375da332a4b5b56d605b0068684c) | +955/−0 |
| clickhouse | [flux-clickhouse/src/lib.rs](https://github.com/gattaca-com/flux/pull/176/files#diff-bd8c75a46e3e49298e160af6db5cdce07ae278d5288d189ff4b2f4bc3e6f2adb) | +11/−3 |
| auth | [scram.rs](https://github.com/gattaca-com/flux/pull/176/files#diff-22fdc35e8e7b70009487de7a6907bd0a8f7610d449ded8d81722bf06a3912d9d) | +125/−0 |
| binary encoder | [copybinary.rs](https://github.com/gattaca-com/flux/pull/176/files#diff-34438099c872425bbf7cd8e3c9a831314517a3d87adafbf03ea05dbc8d808b00) | +342/−0 |
| text encoder | [copytext.rs](https://github.com/gattaca-com/flux/pull/176/files#diff-3244cb28b437fed6e76dc6b4e1dc29d0862959adc087af2107c1eaecc3da24c7) | +307/−0 |

<details>
<summary>Supporting: root, tests, flux-postgres (+812/−104)</summary>

| area | file | lines |
|---|---|---|
| root | [Cargo.lock](https://github.com/gattaca-com/flux/pull/176/files#diff-13ee4b2252c9e516a0547f2891aa2105c3ca71c6d7a1e682c69be97998dfc87e) | +14/−0 |
| root | [Cargo.toml](https://github.com/gattaca-com/flux/pull/176/files#diff-2e9d962a08321605940b5a657135052fbcef87b5e360662bb527c96d9a615542) | +5/−0 |
| tests | [clickhouse.rs](https://github.com/gattaca-com/flux/pull/176/files#diff-13f2651ed7935ffac4edec88589935fc6c7c99ff9f63324adf6c190d20eb47cc) | +22/−7 |
| tests | [tests/http.rs](https://github.com/gattaca-com/flux/pull/176/files#diff-23ffccc1141a4bdfc83cb541e2b7a08d8bcb017f41b6239c61710415cc7b77e1) | +138/−97 |
| flux-postgres | [flux-postgres/Cargo.toml](https://github.com/gattaca-com/flux/pull/176/files#diff-35a3f29c02f0c0a21ecfaf98ab634f93b3e537a761f2b70da425b89a8516eb2d) | +23/−0 |
| tests | [postgres.rs](https://github.com/gattaca-com/flux/pull/176/files#diff-6609cda7ec5314ef142265fcdd09974b142567be305cdbf8303f499ea8be5e9a) | +610/−0 |
</details>


louisponet:
Review of `da874a7` (elegance and performance focus; second pass after the queue commit).

## What improved
- `query`/`copy`/`copy_rows` return a `QueryId` immediately; no caller-side retry loop. ✔
- `copy` takes `Vec<u8>` by value (one copy gone; two remain: `outbox` → network buffer). ◐
- `on_event(&event) -> bool` is now a plain filter; outcomes come out of `drive`. The callback-in-callback is gone. ✔
- The caller-owned `TcpNetwork` layering is argued explicitly (one epoll per tile). Accepted.

## New problems — these outrank everything else

**1. `copy` resends in-flight COPYs, which duplicates rows.** [`RETRIES = 3` is hard-coded for `copy`](https://github.com/gattaca-com/flux/blob/da874a7c638dca92e95576b06dd06633e8ace6c7/crates/flux-postgres/src/lib.rs#L317); the [`Disconnected` arm](https://github.com/gattaca-com/flux/blob/da874a7c638dca92e95576b06dd06633e8ace6c7/crates/flux-postgres/src/lib.rs#L510) requeues whatever was in `State::Copy` without knowing whether the server committed. COPY FROM STDIN in a simple query autocommits when `CopyDone` is processed; lose the socket between the server's commit and our `C`/`Z`, and the resend inserts the whole batch again. For append-only telemetry tables with no unique key that is silent duplicates at exactly the moment a hiccup happens. The doc admits it: *"the server may see it more than once."*

The queue already gives the safe half for free: requests still *in the queue* never left the client and are trivially retried; requests *in flight* are the ones that must fail with `Disconnected` and let the owner of the data decide. `HttpNetwork` makes this a per-request `retries` argument because ClickHouse dedups identical insert blocks and S3 PUTs are idempotent; Postgres COPY is neither. Drop `RETRIES`, `retries_left`, and the requeue on disconnect.

**2. [`insert.rs`](https://github.com/gattaca-com/flux/blob/da874a7c638dca92e95576b06dd06633e8ace6c7/crates/flux-postgres/src/insert.rs) (325 lines) exists to paper over (1).** It is a second serde `Serializer`, ~250 lines near-identical to `copybinary`, rendering rows as escaped `E'…'` text literals in one giant `INSERT … VALUES` so that `ON CONFLICT DO NOTHING` makes the resend "safe". But it is only idempotent if the table happens to have a unique index covering the row (without one `DO NOTHING` is a no-op and the duplicates return); it is the slow path server-side (parse a multi-MiB statement, text→binary per value) — the opposite of what a telemetry pipeline wants; and hand-rolled SQL literal escaping is a surface nobody asked for. Nobody requested INSERT. Delete `insert.rs`, `insert_rows`, `OnConflict`, and the two tests. `Bytea` is fine to keep — `copybinary` needs it too.

**3. Full-queue eviction inverts the contract.** [`enqueue`](https://github.com/gattaca-com/flux/blob/da874a7c638dca92e95576b06dd06633e8ace6c7/crates/flux-postgres/src/lib.rs#L355) evicts the *oldest* accepted requests as `Error::Dropped`, so an id the caller was told is queued fails later. `HttpNetwork` refuses the *new* request and hands the body back (`Result<RequestId, Vec<u8>>`): the caller keeps control, nothing accepted is lost, and there is no eviction loop or `Dropped` variant. Match it.

## Still open from the first pass
- `TcpGroupConfig::on_connect_msg` is exactly the static startup packet; [`queue_startup` in `Connected`](https://github.com/gattaca-com/flux/blob/da874a7c638dca92e95576b06dd06633e8ace6c7/crates/flux-postgres/src/lib.rs#L483) goes away.
- `max_frame_size: usize::MAX` on the Raw group, as `HttpNetwork` does; then [`SEND_CHUNK_BYTES`](https://github.com/gattaca-com/flux/blob/da874a7c638dca92e95576b06dd06633e8ace6c7/crates/flux-postgres/src/lib.rs#L32) and the `out_sent` chunk loop go.
- Pipeline `Q`+`d`+`c` in one send at dispatch time — the backend accepts-and-ignores `CopyData`/`CopyDone` outside copy-in mode — instead of [parking `d`/`c` until `G` arrives](https://github.com/gattaca-com/flux/blob/da874a7c638dca92e95576b06dd06633e8ace6c7/crates/flux-postgres/src/lib.rs#L823) on a *later* `drive`: removes `CopyOut.sent`, the deferred send, a round trip, and one poll-loop latency per batch. Verify once against a real server.
- `Postgres::dispatch_inner/startup/auth/query_msg/copy_msg` remain 5–7-arg associated fns operating on `State`/`Startup`/`QueryOut`/`CopyOut`; repo rule says methods on those types.
- MD5 auth is deprecated and brings `md-5` plus its machete exemption.
- Tests: 12 now (was 8), 846 lines. Target: one fake-server e2e (the byte-at-a-time one is the keeper) plus a manual live SCRAM check.

Net: the diff should be shrinking toward ~1000 lines (`lib.rs` + `copybinary` + `scram` + one test); it is at 2632 and growing. Fix (1) before anything else — it is a data-integrity bug shipped as a feature.

louisponet:
Addressed in 67761e9:

1. In-flight resends dropped — disconnect always reports `Disconnected`. `RETRIES`, `retries_left`, and requeue-on-disconnect are gone (unsent requests still go back to the queue head on a failed dispatch).
2. `insert.rs`, `insert_rows`, `OnConflict`, and both tests deleted. `Bytea` kept.
3. Refuse-new: `query`/`copy_rows` return `Option` (`None` when full; `copy_rows` also panics on unencodable rows, `None` on empty), `copy` hands its `Vec<u8>` back. No `Dropped`, no eviction loop.
4. Startup via `on_connect_msg` (credential/database/parameter builders assert pre-drive config); group uses `max_frame_size: usize::MAX`, chunk loop and `out_sent` gone.
5. `Q`+`d`+`c` pipelined in one zero-copy send at dispatch. Verified against PG 18 `postgres.c`: stray `CopyData`/`CopyDone`/`CopyFail` are accepted-and-ignored, so a rejected COPY still surfaces its `ErrorResponse` cleanly. No live server available in this environment, so a live check remains worthwhile.
6. Protocol handlers are now methods on `State`/`Startup`/`QueryOut`/`CopyOut`.
7. MD5 dropped (code, both manifest entries, machete exemption, test); servers asking for it get `UnsupportedAuth("MD5")`.
8. Tests: the byte-at-a-time keeper only (353 lines incl. harness). Total is now ~1700 (`lib` 894 + `copybinary` 317 + `scram` 125 + tests 353); the remainder is the state machine, parsers, and auth.

bartoszmodelski (approved):

